PR #27, opened 2020-05-09
Update docs of myNode install guide for Raspiblitz

andoniabedul:
The guide it's pretty much the same thing, only few things of paths and configurations change.


Evanfeenstra:
great, thanks!

PR #36, opened 2020-05-25
Fixed count now done without retrieving all the rows and counting them

gonzaloaune:
@Evanfeenstra this fixes an issue where the CPU will climb to > 100% if you have one too many contacts / messages.


Evanfeenstra:
nice!

PR #41, opened 2020-05-30
Removed some unneeded steps and mistakes.

CodingMuziekwijk:
Removed some unneeded steps and mistakes.


dennisreimann:
This seems outdated as the install guides for the individual platforms have been split.

PR #42, opened 2020-06-01
Bugfix sqlite sequelize versions

gonzaloaune:


package.json
```diff
@@ -91,7 +91,7 @@
     "node-sass": "^4.14.1",
     "nodemon": "^2.0.1",
     "sass-loader": "^8.0.0",
-    "sqlite3": "^4.2.0",
+    "sqlite3": "4.1.1",
```
gonzaloaune, line 94: Is this going to affect anything @Evanfeenstra? We can't really compile again in the nodes, do you have any idea how would this work?

Evanfeenstra: i would be very surprised if it caused any issue, should be fine

PR #134, opened 2020-12-21
Add updates for Voltage Integration

gkrizek:
Couple things that are needed to work with Voltage nodes. We need to support more cipher suites for TLS certificates and also explicitly add in the configured LND certificate into Node's CA pool.

I also removed the console.logs that logged the password and encrypted macaroon data. I don't think those should probably be kept in logs.

Thanks!


Evanfeenstra:
thanks for cleaning up those logs

PR #181, opened 2021-04-09
Add optional router and signer macaroon locations

dennisreimann:
Closes #177.

I didn't find any tests, let me know if I missed something.

PR #182, opened 2021-04-09
Upgrade TypeScript; fix build

dennisreimann:
Working on #181 I noticed the build is broken due to an old TypeScript version.


dennisreimann:
@Evanfeenstra What is the state of the build? I saw that the [tsc package is deprecated](https://www.npmjs.com/package/tsc) and the build dos not work with the regular TypeScript version. I'd like to help, but am unsure how to proceed :)

AaronDewes:
>  I saw that the tsc package is deprecated and the build dos not work with the regular TypeScript version. I'd like to help, but am unsure how to proceed :)

Very likely, the tsc package was a mistake in the first place, because `tsc` is provided by `typescript`.

PR #207, opened 2021-07-07
adds linting

bucko13:
This PR adds eslint and prettier and runs it on the repo's files. 

Rules can be found in the .prettierc and .eslintrc files. Feel free to request any changes!

PR #210, opened 2021-07-22
adds eslint, prettier, and pretty-quick pre-commit hook via husky

bucko13:
overrides #207 

Obviously this touches a lot of files and I didn't test or check very closely (still getting used to the codebase). The `tsc` build passed without any problems though. I tried to run the cluster-test.js and ran into an error `Cannot use import statement outside a module` but it looks like I'm also getting that on master so hopefully unrelated (though it's possible I've done something on my local machine)


Evanfeenstra:
Thanks Buck!

PR #227, opened 2021-09-13
adds lsat controllers and testing utilities

bucko13:
Changes include:

* new lsat model for saving lsat data in the database
* save, get, list, update, and delete endpoints for managing lsats
* various edits to comply with typescript and linter requirements across some older files that were touched
* integration-style testing framework (helper utilities adapted from electron app) for testing relay controllers
* integration-style tests for new lsat controllers
* updates testing readme for sphinx-stack and integration tests
* remove dist directory from git history


Evanfeenstra:
Looks really great @bucko13! The only thing I would request is to add `dist` back in there. Not needing to build every time speeds up updating our cloud instances, also home node users are accustomed to this.

PR #228, opened 2021-09-15
Asdfasdf

Evanfeenstra:
- fix proxy hub pings


jimbojw:
Nit: Descriptive comments on PRs and commits would help others understand the changes both in the present and for the historical record.

PR #232, opened 2021-10-12
Fixing docker build error messages

kevkevinpal:
added these two lines because we were getting an error message of ERR! enoent undefined ls-remote -h -t https://github.com/tierion/js-macaroon.git which meant that our git binary was not there so I added apt-get git and then also an npm error which complained about the package.json not being available and it seemed our own package-lock.json was interfering with that.

 

- [ ] Test it on your own machine before merging into master please


kevkevinpal:
Closing this PR because it seems we got the changes in master now

PR #240, opened 2021-10-29
Boost and Tribe Images testing

kevkevinpal:
I made changes to migrate over the 
`src/tests/controllers/tribeImages.test.ts` and `src/tests/controllers/boostPayment.test.ts` files

I made some changes to use the external_ip value in a few functions but I also made a change in 
`src/tests/utils/msg/sendTribeMessageAndCheckDecryption.ts` to encrypt a message to the tribe group key instead of with the second node contact key because the message content at the end was not matching. 

Also added the `src/tests/utils/msg/sendBoost.ts` file to be able to send boosts now

will open a draft PR for now to give the code another look at and try to clean it up a bit

This should also fix the `tribe.test.ts` file now as I think it was failing before because of the issue in the `sendTribeMessageAndCheckDecryption` file the same way


Evanfeenstra:
awesome!

PR #246, opened 2021-11-28
Update myNode_deployment.md

bucko13:
Some suggested updates for things I ran into when going through the setup:

* nginx config was set to listen on the same port as sphinx-relay which would result in a process conflict on that port. Updated docs to have nginx listen on 54001 and forward to sphinx on 53001
* added some extra details to help explain how the routing is working
* some hints for possible errors from acme.sh when setting up ssl serts


kevkevinpal (approved):
This looks good to me, thanks for including additional documentation!

PR #249, opened 2021-12-10
converted the files but when running seems to be failing when doing b…

kevkevinpal:
So I added this chatPaymentTest it seems to be working for
alice-> bob
alice-> Carol
bob-> Alice
but then it fails on bob-> Carol and is giving 
`failed to getContactAndCheckKeyExchange`

So I think I'll continue to look into that, but for the most part it seems the test itself is working fine


kevkevinpal:
So I think I found the issue in this commit dc6dfd6

steps to reproduce
build new relay container with changes from new commit
`npm run build`
`docker build -t <newName> .`

get sphinx-stack up
go to sphinx stack dir
go into `docker-compose.yml` and then switch `alice, bob, carols` container image to the new one you made
`./clearall.sh`
`docker-compose up -d`

now monitor tills sphinx-stack_lndsetup_1 and `sphinx-stack_relaysetup_1` state turns to `exit`

now go back to the relay dir and change the `./src/tests/configs/nodes.json` to the one you currently have in the stack repo in `./relay/NODES.json`
now you can run
`npx ava src/tests/controllers/chatPayment.test.ts --verbose --serial --timeout=2m`
and you should see it fail with the old code and if you repeat the steps with the new code it should pass

PR #251, opened 2021-12-13
Migrating the 3tribe messages test over

kevkevinpal:
Migrating 3 tribe messages test over

PR #252, opened 2021-12-14
Tribe3Escrow test

kevkevinpal:
adding the tribe3escrow test from win/linux repo


kevkevinpal:
This PR is depenendent on https://github.com/stakwork/sphinx-relay/pull/251 so if we can merge that and then merge this one that would work, since the diff is including the changes on that PR

PR #255, opened 2021-12-15
adding in editTribe test

kevkevinpal:
edit tribe test and some helper functions added, looks to be working fine on sphinx-stack

PR #256, opened 2021-12-15
Stream payment tests

kevkevinpal:
adding stream payment tests


kevkevinpal:
Dependent on
https://github.com/stakwork/sphinx-relay/pull/255

because I added the config changes in that branch and this one needed them to compile on my side

PR #267, opened 2021-12-28
Automated testing for relay

kevkevinpal:
I was testing on my fork so some changes may need to be made but the automated testing was passing with one test there, Still need to make final changes but we have one test passing now


kevkevinpal:
Also when we merge this branch we may want to do a squash and merge since there were over 100 commits since I needed to test on github

kevkevinpal:
Also, this PR includes the issue_closed/issue_opened yml files I can remove them if we don't want them in this PR and in a separate one instead

Issue #271, opened 2022-01-12
Restore improvements


tomastiminskas:
In order to improve the restore process and allow users to see progress and stop restore in the middle we need some changes on relay endpoints:

### THESE CHANGES SHOULD APPLY TO THE TWO MESSAGES ENDPOINT (`messages.ts`):
```
- getAllMessages
- getMsgs
```

### CHANGES

- Add `new_messages_total` key on response. This should return the total amount of messages (not the total on the current page, but the total count). That way the app can calculate the progress based on total and current page. 
- Change messages order from `['id', 'asc']` to `['id', 'desc']`. So we can implement a way to allow user to stop restore process in the middle. Right now the oldest messages are returned first, then if stopped newest messages won't be restored

kevkevinpal:
@tomastiminskas let me know if this PR looks about what you were looking for, also will add some more checks for the test added in this PR

kevkevinpal:
@tomastiminskas are we good to close this issue?

tomastiminskas:
@kevkevinpal yes!

PR #272, opened 2022-01-13
Restore improvements

kevkevinpal:
made changes in regard to this issue https://github.com/stakwork/sphinx-relay/issues/271

reverses the order that `getAllMessages` and `getMsgs` returns the `new_messages` and also adds a `new_messages_total` field that is the count of the the `new_messages`

Added some tests for this but still need to add some more, also updated test file for `src/tests/utils/get/getCheckBotMsg.ts` since it used to look for the last value but now we swapped it to be the first value in the array


tomastiminskas:
@kevkevinpal on both endpoint you are returning the length of the messages in the current page, which will be always equal to the "limit" param coming from app. What we would need to add is the total of the messages with the same where clause, but without the limit and the offset (so I guess we need to add a count query to the db). At least that's the best way I could think of to calculate the progress of the restore process on the app side. Let me know if you have a better idea

kevkevinpal:
@tomastiminskas in 2dcb561 I used the count method on the model before we add the limit and offset clause, let me know if this looks good or if you have any concerns.

I also updated the test in 75e0559 to send 4 messages but only retrieve 2 messages

also not too sure what offset is supposed to be so I just set that as 0

tomastiminskas:
@kevkevinpal that looks good but you need to apply the same count logic to the `getAllMessages` endpoint, since that is used when date param is not present on the call. Once that's done I can ask Evan to deploy the branch to one of my nodes and test there. 

Not sure what do you mean with this:
"also not too sure what offset is supposed to be so I just set that as 0"

Can you clarify or point me to the code where you did that?

kevkevinpal:
> @kevkevinpal that looks good but you need to apply the same count logic to the `getAllMessages` endpoint, since that is used when date param is not present on the call. Once that's done I can ask Evan to deploy the branch to one of my nodes and test there.
> 
> Not sure what do you mean with this: "also not too sure what offset is supposed to be so I just set that as 0"
> 
> Can you clarify or point me to the code where you did that?

yea for the offset question it was in the test, not sure if I should be testing different offsets or if just setting it at 0 is fine for this test
https://github.com/stakwork/sphinx-relay/pull/272/files#diff-4dd99ff06ec63eed5c9f9b640091fccd3beb9cdbf52abcbd06f7c7e67abe36caR36

for the getAllMessages would I be able to do something similar but just with no date?
I can push a commit soon

tomastiminskas:
@kevkevinpal yeah testing with offset = 0 is fine. 

Regarding the getAllMessages changes, definitely apply the same logic with applied on getMsgs but with no date. Once It's pushed I will check and if it looks ok we can deploy to one of my nodes to test. Thanks

kevkevinpal:
I updated the getAllMessages endpoint in this commit 
58a9444

and added some additional testing in these commits
b7a0f54
6120bae

should be good to go @Evanfeenstra would you be able to take a look and see if it looks good on your end?

tomastiminskas:
@Evanfeenstra if it looks good to you, could you please deploy to Tom Sim and let me know so I can test there? Thanks in advance

tomastiminskas:
@kevkevinpal @Evanfeenstra I was doing the kotlin changes for these restore improvements and found some backward compatibility issues with the order change we applied on the 2 messages endpoint. Basically because before these changes the most recent message was the last one coming from relay and app was just updating the chat last message every time a message comes in. So with the new changes, the last message coming from relay is actually the oldest one, so app is now showing the oldest message on the chats list rows. 

What we need to do is: 
- Keep the old order by default on both messages endpoints ("asc")
- Receive a new order param that will override the default order value. So new apps will send `order: "desc"` and that will make the endpoint to apply the new order

This way we can keep old apps getting the messages on the order they were receiving it, and new apps sending the new param so messages come in the inverse order.

Let me know your thoughts

kevkevinpal:
@tomastiminskas looks like Evan already merged changes 

https://github.com/stakwork/sphinx-relay/commit/cf88a5c5c75cf10bce8d2d376bd6a3cef1fae13e

tomastiminskas:
@kevkevinpal yes, the changes are done. Thanks

PR #276, opened 2022-01-26
Replay attacks

kevkevinpal:
This is a change to stop using the `x-user-token` in the header and instead supplying the client an `transportToken` to then pass a new header `x-transport-token` which will be comprised of the `encrypt(transportToken, `${authToken}|${timestamp}`

This change does not include a db to stop repeated `x-transport-tokens` being used within the 1 minute timeframe set, We will need to include this in another PR with more tests

still some changes need to be made to cleanup


kevkevinpal:
this PR rely's on https://github.com/stakwork/sphinx-stack/pull/15

so we'll need to merge the above first and change the github workflow file to test properly

kevkevinpal:
added these commits to fix some tests after the flag for the `desc` and `asc` flags 
65a5e97 57ee5cc

kevkevinpal:
I'm going to add a test for a minute old request to see if that fails

kevkevinpal:
Test for old transport token 
d874f59

src/auth.ts
```diff
@@ -82,13 +83,58 @@ export async function ownerMiddleware(req, res, next) {
...
+    // The token will be the first item
+    token = splitTransportToken[0]
+
+    // The second item will be the timestamp
+    const splitTransportTokenTimestamp = splitTransportToken[1]
+
+    // Check if the timestamp is within the timeframe we
+    // choose (1 minute here) to clear out the db of saved recent requests
+    if (
+      new Date(splitTransportTokenTimestamp) < new Date(Date.now() - 1 * 60000)
```
kevkevinpal, line 126: Should I make this available as a config?

src/utils/config.ts
```diff
@@ -108,6 +108,10 @@ export function loadConfig() {
     allow_test_clearing: ENV.ALLOW_TEST_CLEARING || config.allow_test_clearing,
     sql_log: ENV.SQL_LOG || config.sql_log,
     dont_ping_hub: ENV.DONT_PING_HUB || config.dont_ping_hub,
+    transportPrivateKeyLocation:
+      ENV.TRANSPORT_PRIVATE_KEY_LOCATION || config.transportPrivateKeyLocation,
+    transportPublicKeyLocation:
+      ENV.TRANSPORT_PUBLIC_KEY_LOCATION || config.transportPublicKeyLocation,
```
kevkevinpal, line 114: Should I create a default value for the location of these new tokens

Evanfeenstra: yes, make a default location so existing nodes dont need to update their config file

kevkevinpal: yup updated in these two commits 9bfd2d7 be7ea04

src/auth.ts
kevkevinpal: Added this into the config aswell with a default value
e3c819b

Evanfeenstra:
Some very minor comments:
- https://github.com/kevkevinpal/sphinx-relay/blob/replayAttacks%23269/src/auth.ts#L148. If a token is included without a "|" separator, then `splitTransportToken[1]` will panic and the route will hang. Check that splitTransportToken.length>1 first.
- remove "TODO" comments if they are done https://github.com/kevkevinpal/sphinx-relay/blob/replayAttacks%23269/src/auth.ts#L165
- column names should be snake_case: https://github.com/kevkevinpal/sphinx-relay/blob/replayAttacks%23269/src/utils/migrate.ts#L180. They are automatically converted to camelCased by `sequelize`, so in your actual code you can still use `transportToken` just the same.
- no need to add the column since its in the table: https://github.com/kevkevinpal/sphinx-relay/blob/replayAttacks%23269/src/utils/migrate.ts#L189
- no need to log this error. After the first startup, it will "error" every time it restarts since the table arleady exists: https://github.com/kevkevinpal/sphinx-relay/blob/replayAttacks%23269/src/utils/migrate.ts#L184

PR #280, opened 2022-02-01
add shorter date format for logger

Rassl:


kevkevinpal:
Nice work lemme check to see it's working on my end and then we can send you over the 10,000 sats :) are you on Sphinx? We can pay you with lightning on there

PR #293, opened 2022-02-22
[Security] Updating package.json with newest socet.io version

kevkevinpal:
**Careful this change may break parts of the app that use socket.io we should check to see if they're still working**

When I ran `npm audit` it gave socket.io as a high severity error. It is two major versions updated so I might break some functionality

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "sequelize-typescript": "^1.1.0",
     "short-uuid": "^3.1.1",
     "sjcl": "^1.0.8",
-    "socket.io": "^2.4.0",
+    "socket.io": "^4.4.1",
     "sphinx-bot": "^0.2.21",
     "sqlite3": "4.1.1",
     "tail": "^2.0.3",


kevkevinpal:
Its probably adding a socket.io test to make sure we don't have this break in the future and its also a good test just to have

Evanfeenstra:
yes i would not merge this until its tested, shouldnt be too hard to add a socket.io test.

Backward compat will be a challenge here....

kevkevinpal:
> yes i would not merge this until its tested, shouldnt be too hard to add a socket.io test.
> 
> Backward compat will be a challenge here.... I guess we need to support both versions (on a different route) for a while

ok sounds good when trying to implement the test should I use `socket.io-client` kinda like how we do in win/linux or is there another approach you were thinking of?

kevkevinpal:
closing for now since I have a new branch/PR I'm working on for these changes

https://github.com/stakwork/sphinx-relay/pull/296

PR #303, opened 2022-03-14
Feat/hmac

Evanfeenstra:
- hmac authentication
- opt-in... if client doesn't post an hmac key than the auth isnt used
- each request is signed like `{method}|{path}|{rawBody}` (path starts with "/" and includes url queries)


kevkevinpal:
I think you might need to update this to `sphinxlightning/sphinx-relay:latest` since we're building a new tag for the old name rn
https://github.com/stakwork/sphinx-relay/blob/master/.github/workflows/pull_request.yml#L24

PR #306, opened 2022-03-16
 Code for stakwork/sphinx-stack/issues/21

Lombitron:
- Refactor `signAndSend()` into two methods `sign()` and `post()` 
 - Removed publish to MQTT & replaced with a post to `sphinx-tribes` new batch endpoint.
 - Changed to sign each message, batch it up into a new JSON object, then send that to the new endpoint in `sphinx-tribes`

### DIFF
--- a/src/network/send.ts
+++ b/src/network/send.ts
@@ -1,15 +1,12 @@
 import { models } from '../models'
 import * as LND from '../grpc/lightning'
 import { personalizeMessage, decryptMessage } from '../utils/msg'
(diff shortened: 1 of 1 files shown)


Evanfeenstra:
Run `npm run tsc`, you will see that the code does not compile. You need to keep the `signAndSend` function because its used all over the place.

Also you lost the `LND.keysendMessage` which is the actual lightning message. You can only send MQTT `if (mqttTopic)`, otherwise lightning

PR #287, opened 2022-02-12
Add `eslint` to project and fix problems

antonilol:
closes #286

PR #307, opened 2022-03-16
Eslint utils/tribes.ts

kevkevinpal:
Ok since @antonilol PR
https://github.com/stakwork/sphinx-relay/pull/287

is very large and we're not too confident merging it all at once we're going to break it down to a few PR's to make it more manageable to review and test

This change is specifically for the `./src/utils/tribes.ts` file and you can see us progress in the `.eslintignore` file as we continue to unignoring multiple files as we go along starting with the `utils` folder probably


kevkevinpal:
@Evanfeenstra looks like the tests are passing let me know if the changes look fine or if anything stands out to you

kevkevinpal:
> yeah i made that TribeInterface when i did not know the code enough and did not know those models existed, but looking at it now it fits the Tribe model

So it looks like to might have made that tribe interface because typescript gets mad when using the reserved word `private` but it wouldn't make sense to put two fields in there and have one be `is_private` this should be good to go but @Evanfeenstra let me know if you have any additional comments

Evanfeenstra:
looks good! i'll go ahead and merge

PR #309, opened 2022-03-17
fix random things i encountered while scrolling READMEs

antonilol:


kevkevinpal:
Looks like a fairly straight forward PR LGTM

kevkevinpal:
Update with a 50k sat payout as a thank you @antonilol

PR #312, opened 2022-03-18
eslint for ./src/controller/invoices.ts

kevkevinpal:
https://github.com/stakwork/sphinx-relay/pull/287

This change is specifically for the ./src/controllers/invoices.ts file and you can see us progress in the .eslintignore file as we continue to unignoring multiple files as we go along starting with the utils folder probably

I added a new `RelayRequest` interface that we could use for the Request Type instead


kevkevinpal:
@Evanfeenstra if you checkout [this job](https://github.com/stakwork/sphinx-relay/runs/5603901137?check_suite_focus=true) under the `build and push dist folder` 

you can see it not pushing to the PR, if we give permissions `github-actions[bot]` we can never have to worry about building the dist folder before pushing for a PR

antonilol:
i made that in the eslint pull, and it got merged in #307, but doesn't seem to work not because of permissions, but because there are two workflows, one for automatic eslint changes (eslint automatically changes `let` without reassignment to `const`) and one for the dist folder.
they both start on the same commit, the eslint thing was faster this time, pushed the changes and the other workflow got an error that the local branch is not up to date.

antonilol:
another small remark is that the push done in this eslint workflow edited the package-lock, so i will add a `git add src` there to be sure it only edits that folder

kevkevinpal:
awesome I see the new PR

PR #313, opened 2022-03-18
fix workflows

antonilol:
https://github.com/stakwork/sphinx-relay/pull/312#issuecomment-1072631100


.github/workflows/lint.yml
```diff
@@ -23,4 +20,5 @@ jobs:
           git config --global user.name "Github Actions"
```
kevkevinpal, line 20: just a heads up I updated the `npm run lint` command you might wanna update that aswell

antonilol: > just a heads up I updated the npm run lint command you might wanna update that aswell

if the npm script is changed, this will change as well since it just runs npm run lint

kevkevinpal: yea if you can update it to what you had in your original PR that would be great, think this is it

`"lint": "eslint ./src/**/*.ts ./src/*.ts --fix"`

antonilol: done

kevkevinpal:
Looks like this PR is failing because of an eslint issue that was merged earlier fixing in another PR

antonilol:
ok now it works

antonilol:
> when i tried to run these tests locally a few weeks ago the docker wait command gave an error but there was a NODES.json file in stack.
> and also just found out the nodes.json in relay is commited on git and ignored.
> 
> instead of the docker wait command, maybe you can (or i) add a for loop that checks if NODES.json exists and also times out to prevent tests getting stuck
>
> i can add it to my workflow PR

https://github.com/stakwork/sphinx-relay/pull/314#issuecomment-1073028479

.github/workflows/lint.yml
```diff
@@ -23,4 +20,5 @@ jobs:
...
           git remote add github "https://$GITHUB_ACTOR:$GITHUB_TOKEN@github.com/$GITHUB_REPOSITORY.git"
           git pull github ${GITHUB_REF} --ff-only || echo -n
-          git commit -am "Automatic linter fixes" && git push github HEAD:${GITHUB_REF} || echo -n
+          git add src
```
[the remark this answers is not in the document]
antonilol: true

fixed this by merging them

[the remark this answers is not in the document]
antonilol: this is also much faster, `npm i` takes some time because grpc needs to be compiled, now it has to be compiled one time less

kevkevinpal (approved):
Looking good LGTM @Evanfeenstra any thoughts on these workflows?

Evanfeenstra:
looks good to me!

PR #314, opened 2022-03-18
eslint on network and msg

Evanfeenstra:
- src/network typings
- src/utils/msg typings


src/controllers/messages.ts
```diff
@@ -358,7 +358,7 @@ export const sendMessage = async (req, res) => {
   if (reply_uuid) msgToSend.replyUuid = reply_uuid
   if (parent_id) msgToSend.parentId = parent_id
 
-  const sendMessageParams: { [k: string]: any } = {
+  const sendMessageParams: network.SendMessageParams = {
```
kevkevinpal, line 361: Styling preference but it looks cleaner when its just `SendMessageParams` rather than `network.SendMessageParams`

src/helpers.ts
```diff
@@ -142,7 +142,7 @@ export const performKeysendMessage = async ({
   destination_key: string
   route_hint?: string
   amount: number
-  msg: { [k: string]: any }
+  msg: Partial<network.Msg>
```
kevkevinpal, line 145: Same thing here just a styling preference

src/controllers/messages.ts
Evanfeenstra: fo sho

kevkevinpal (approved):
Looks good minus the two comments

src/types.ts
```diff
@@ -0,0 +1,7 @@
+import * as express from 'express'
+import { Contact } from './models'
+
+export interface Req extends express.Request {
```
antonilol, line 4: did something like this on #287 
https://github.com/stakwork/sphinx-relay/pull/287/files#diff-3e0f247bfd5389158886d6f0bff49895160ba71a8dc3d874a5760f833019e568 (didnt know the type when making this lol)

antonilol: note: diff can take a while to load and scroll down

Evanfeenstra: nice, ya the `req.owner` is added in by the auth middleware, and is used in pretty much every route, so wanted that to be strongly typed

kevkevinpal:
@Evanfeenstra so it looks like after your first commit on this PR the tests start stalling and NODES.json is never generated in the tests

So something might of broke it in those newer commits

antonilol:
when i tried to run these tests locally a few weeks ago the docker wait command gave an error but there was a NODES.json file in stack.
and also just found out the nodes.json in relay is commited on git and ignored. 

instead of the docker wait command, maybe you can (or i) add a for loop that checks if NODES.json exists and also times out to prevent tests getting stuck

src/controllers/chatTribes.ts
```diff
@@ -509,7 +509,7 @@ export async function receiveMemberApprove(payload) {
 
   const amount = chat.priceToJoin || 0
   const theChat = chat.dataValues || chat
-  const theOwner = owner.dataValues || owner
```
kevkevinpal, line 512: So I'm not sure if this line is causing the tests to not pass but I'm not sure. This seems the most suspect though

Evanfeenstra: finally figured it out, it was an issue in app.ts. Should be good now, lets see if all the tests pass

PR #320, opened 2022-03-21
add linux install doc

### LINKED ISSUE #318: Document setup for LND/Sphinx on linux box $@250,000
prompted by a question from Yidir on telegram

antonilol:
closes #318

TODO
- [x] add apache2 web server ssl proxy setup (i use it myself)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ If you are using the Docker file, the port is `3300` ([1](https://github.com/sta
 ## Deployment
 
 - [Docker deployment](docs/docker_deployment.md)
+- [Linux deployment](docs/linux_deployment.md)
 - [Raspberry Pi/myNode deployment](docs/myNode_deployment.md)
 - [Raspberry Pi/Raspiblitz deployment](docs/raspiblitz_deployment.md)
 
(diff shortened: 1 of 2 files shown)


docs/linux_deployment.md
```diff
@@ -0,0 +1,515 @@
+# Deployment on Linux
+
+This guide is focused on installing Sphinx-relay on top of LND on Linux.
+
+### Preparations
+
+- Connect to your node via SSH to the user account where your LND runs (most people use `bitcoin`)
+
+```sh
+$ ssh bitcoin@ip.ad.dr.ess
```
kevkevinpal, line 10: Is this the actual ip address? It might be worth noting to the person reading this that they might need to find this info and where they can find that

README.md
```diff
@@ -111,6 +111,7 @@ If you are using the Docker file, the port is `3300` ([1](https://github.com/sta
 ## Deployment
 
 - [Docker deployment](docs/docker_deployment.md)
+- [Linux deployment](docs/linux_deployment.md)
```
kevkevinpal, line 114: This is for clearnet only right? I'm assuming you didn't use tor. Do you know how much more different it is if someone wanted to use tor instead?

antonilol: > This is for clearnet only right?

yes

> Do you know how much more different it is if someone wanted to use tor instead?

no, actually didnt use tor before with proxy (i do have tor browser bc its easy to use)

kevkevinpal: ok sounds good, guide is looking great btw :)

docs/linux_deployment.md
antonilol: done

kevkevinpal (approved):
Haven't done this myself but the guide is looking good and thorough

PR #326, opened 2022-03-23
eslint for ./src/controller/chats.ts and some to contacts

kevkevinpal:
updated eslint rc to give warnings rather than ignoring rc rules
updated chats file for eslint
updated contacts file aswell for eslint


src/controllers/chats.ts
```diff
@@ -205,10 +229,10 @@ export async function createGroupChat(req: Req, res) {
     }
   })
 
-  let chatParams: any = null
```
kevkevinpal, line 208: Is it ok to remove the null here?

antonilol: i try to avoid null as much as possible and replace it with undefined. i would set the type to `Chat | undefined` if it could be undefined, but that's not the case here. `chatParams` in never used where it could be undefined: after the if block you know it is not undefined because it is assigned a value in both possible paths (and before the if it is not used)

Evanfeenstra: ya so probably fine just to do `let chatParam;`?

src/controllers/contacts.ts
```diff
@@ -650,7 +683,7 @@ export const getLatestContacts = async (
   try {
     const dateToReturn = decodeURI(req.query.date as string)
     const local = moment.utc(dateToReturn).local().toDate()
-    const where: { [k: string]: any } = {
+    const where: { tenant: number; updatedAt: { [Op.gte]: Date } } = {
```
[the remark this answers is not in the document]
kevkevinpal: nvm I think there might be too many differences between the where clauses but it seems like a good amount of them use `tenant`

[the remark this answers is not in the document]
Evanfeenstra: Literally basically every single query should have `tenant` in it, because relay can also run in multitenant mode where there are many "owner" contacts

PR #327, opened 2022-03-23
Eslint removing ignore and setting max warning count

kevkevinpal:
It seemed a bit tedious to do everysingle file and then mark it off as complete so what we can do instead is fix all these pretty simple eslint errors that were mostly solved with the `--fix` option in `npm run lint` and then whenever we fix a warning we updated 

`package.json` for the script `lint` to have less warnings

Also this has commits from this PR https://github.com/stakwork/sphinx-relay/pull/326 so it'll be less code in `chats.ts` after it gets merged in


package.json
```diff
@@ -16,22 +16,15 @@
     "dev": "rm -rf dist/ && tsc && cp -r public dist/public && node --inspect ./dist/app.js",
     "build-binaries": "rm -rf dist/ && tsc && cp config/app.json dist/config/app.json && cp config/config.json dist/config/config.json && pkg . --target node12-alpine-x64,node12-macos-x64 --out-path binaries",
     "start-ecs": "cp config/app.json dist/config/app.json && cp config/config.json dist/config/config.json && export NODE_ENV=production && node --max-old-space-size=128 dist/app.js",
-    "lint": "eslint ./src/*.ts ./src/**/*.ts --fix"
+    "lint": "eslint ./src/*.ts ./src/**/*.ts --fix --max-warnings 546"
```
kevkevinpal, line 19: Here is where we have the max warning count that we'll decrement after we fix warnings

.eslintrc
```diff
@@ -5,12 +5,12 @@
...
-    "@typescript-eslint/no-empty-function": 0,
-    "@typescript-eslint/no-unused-vars": 0,
-    "@typescript-eslint/no-explicit-any": 0,
-    "@typescript-eslint/no-non-null-assertion": 0
+    "no-async-promise-executor": 1,
+    "no-constant-condition": 1,
+    "@typescript-eslint/no-empty-function": 1,
+    "@typescript-eslint/no-unused-vars": 1,
+    "@typescript-eslint/no-explicit-any": 1,
+    "@typescript-eslint/no-non-null-assertion": 1,
```
kevkevinpal, line 13: These are the things adding the warnings after we get rid of the warnings we can then set this to `2` so they'll start throwing errors

We can also add or remove these options if we'd like

src/utils/ldat.ts
```diff
@@ -177,9 +177,9 @@ async function testLDAT() {
 }
 
 function serializeMeta(obj) {
-  var str: string[] = []
-  for (var p in obj) {
-    if (obj.hasOwnProperty(p)) {
+  const str: string[] = []
+  for (const p in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, p)) {
```
kevkevinpal, line 182: @Evanfeenstra eslint was erroring on this let me know if this change looks good, one of the few odd ones

src/utils/nodeinfo.ts
```diff
@@ -15,7 +15,15 @@ export enum NodeType {
...
-export function proxynodeinfo(pk: string): Promise<Object> {
+export function proxynodeinfo(pk: string): Promise<{
+  pubkey: string
+  number_channels: number
+  open_channel_data: interfaces.Channel[]
+  largest_local_balance: number
+  largest_remote_balance: number
+  total_local_balance: number
+  node_type: NodeType.NODE_VIRTUAL
+}> {
```
kevkevinpal, line 26: was giving an error for this type this is also one of the few odd ones

src/utils/ldat.ts
antonilol: in my PR i worked around this with 
```diff
@@ -176,15 +176,13 @@ async function testLDAT() {
   sphinxLogger.info(parseLDAT(token2))
 }
 
-function serializeMeta(obj) {
-  var str: string[] = []
-  for (var p in obj) {
-    if (obj.hasOwnProperty(p)) {
-      str.push(encodeURIComponent(p) + '=' + encodeURIComponent(obj[p]))
-    }
-  }
-  str.sort((a, b) => (a > b ? 1 : -1))
-  return str.join('&')
+function serializeMeta(obj: { [k: string]: any }) {
+  return Object.entries(obj)
+    .map(p => {
+      return `${encodeURIComponent(p[0])}=${encodeURIComponent(p[1])}`
+    })
+    .sort((a, b) => (a > b ? 1 : -1))
+    .join('&')
 }
 
 function deserializeMeta(str) {
```

basically using Object.entries instead of a for loop

src/utils/nodeinfo.ts
Evanfeenstra: https://github.com/kevkevinpal/sphinx-relay/pull/14 @kevkevinpal

kevkevinpal:
Merging in this change since there seems to be no major issues and they're all fairly simple changes

PR #328, opened 2022-03-23
fix balance calculation

antonilol:
the total reserve is subtracted from the total balance. this will only work if you know for every channel the sats are in reserve.
this is not the case with newly opened channels. when a channel was opened to me once, the balance in sphinx dropped (it seemed because of this)

this pr calculates total spendable balance per channel so reserve of one channel doesn't affect the other(s)


Evanfeenstra:
Thanks for this @antonilol!

PR #329, opened 2022-03-24
Transport token unix time

kevkevinpal:
updated code because we were using the javascript Date object to make a timestamp but this PR makes it so that we are using the Unix timestamp instead.

I've also updated the corresponding tests to reflect us using the UNIX timestamp as a client

we make the timestamp by using date.now()

switched to this branch cause I started to work off this one now


kevkevinpal:
@Evanfeenstra let me know if this looks good, Pushed new changes for stack to work with the new transport token endpoint name and response variable name

Evanfeenstra:
@kevkevinpal unix timestamp is a 10-digit number, its seconds rather than milliseconds like Date.now(). So i think `moment().unix() - config.length_of_time_for_transport_token_clear * 60000` will not work?? should it be just `* 60` instead? 

It would be best if there was a test to test reusing the token in a short timeframe

kevkevinpal:
> @kevkevinpal unix timestamp is a 10-digit number, its seconds rather than milliseconds like Date.now(). So i think `moment().unix() - config.length_of_time_for_transport_token_clear * 60000` will not work?? should it be just `* 60` instead?
> 
> It would be best if there was a test to test reusing the token in a short timeframe

I do have a test for that but the test had to change because we switched to UNIX time so it changed the input

https://github.com/stakwork/sphinx-relay/pull/329/files#diff-52bf81e673b2af78edbf8e9471dc9fea237c9b879e4835024b4c49d180387716L81

but I can update for it to use that, this must be giving lots of time then

PR #333, opened 2022-03-28
Adding some docs for transport token stuff

kevkevinpal:
will add a bit more but to test it out get the raw openapi.yml file and then navigate here

https://petstore.swagger.io/

and replace with the raw URL


kevkevinpal:
going to merge this even if they're not fully complete

Issue #334, opened 2022-03-29
Lowering max eslint warning count - $@2,000 per warning


kevkevinpal:
In this PR https://github.com/stakwork/sphinx-relay/pull/327 we added a max warning count to the `npm run lint` command
what we want to do is to bring it down as it is 532 as of writing this issue

`How to reproduce`
simply run `npm run lint` and it should tell you how many errors there are

`validation`
change the `lint` script in `package.json` to how ever many you lowered it by and then when you try to run `npm run lint` it should not fail

NOTE: some of the code that isn't tested we should avoid till we have tests for them just to avoid unnecessary breakage you can find the tests that we have automated tests for here

https://github.com/stakwork/sphinx-relay/tree/master/src/tests/controllers

also there is a PR up here that has some of the changes we wanted but this was a very large PR so we wanted to split it up
https://github.com/stakwork/sphinx-relay/pull/287

kevkevinpal:
Look at the above PR ^^^^ for an example on how to make a change

antonilol:
i had the errors and warning at 0 but now after merging master all the time in #287 errors came back and tests failed all the time

have `206 problems (4 errors, 202 warnings)` atm

kevkevinpal:
closing issue

PR #335, opened 2022-03-29
Eslint small change to show how to lower eslint warning count

kevkevinpal:
linking this here since someone might need it
https://github.com/stakwork/sphinx-relay/issues/334


kevkevinpal:
Merging since really small and really just a sample PR how to lower eslint warn count

PR #338, opened 2022-03-31
eslint: selected some parts of #287

antonilol:
i did not build intentionally do keep the diff clean, and we have automatic build after all

### DIFF
--- a/src/builtin/mother.ts
+++ b/src/builtin/mother.ts
@@ -30,7 +30,7 @@ const builtInBotNames = {
   git: 'GitBot',
 }
 
-export function init() {
+export function init(): void {
   const client = new Sphinx.Client()
   client.login('_', finalAction)
 
@@ -45,7 +45,7 @@ export function init() {
     if (!isAdmin) return
 
     switch (cmd) {
-      case 'install':
+      case 'install': {
         if (arr.length < 3) return
         const botName = arr[2]
 
@@ -106,8 +106,8 @@ export function init() {
           }
         }
         return true
-
-      case 'uninstall':
+      }
+      case 'uninstall': {
         if (arr.length < 3) return
         const botName2 = arr[2]
         const chat2 = await getTribeOwnersChatByUUID(message.channel.id)
@@ -132,8 +132,8 @@ export function init() {
             .setDescription('Cant find a bot by that name')
           return message.channel.send({ embed })
         }
-
-      case 'search':
+      }
+      case 'search': {
         if (arr.length < 2) return
         const query = arr[2]
         const bots = await searchBots(query)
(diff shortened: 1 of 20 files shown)


antonilol:
fixed all errors and rebased in a single commit

antonilol:
@kevkevinpal @Evanfeenstra

i picked this one up from more than a month ago and fixed the errors from `npm run build`. can you guys review it?

src/grpc/lightning.ts
```diff
@@ -171,8 +171,8 @@ export async function queryRoute(
       routes: [],
     }
   }
-  return new Promise(async function (resolve, reject) {
-    const lightning = await loadLightning(true, ownerPubkey) // try proxy
+  const lightning = await loadLightning(true, ownerPubkey) // try proxy
+  return new Promise((resolve, reject) => {
```
kevkevinpal, line 175: I'm confused about what swapping these two lines does was it something eslint was throwing a warning about?

Also what does removing the async from the function do, does it affect functionality

src/builtin/mother.ts
```diff
@@ -30,7 +30,7 @@ const builtInBotNames = {
   git: 'GitBot',
```
kevkevinpal, line 30: Again this `src/builtin` folder doesnt have any test coverage yet so it would be ideal to add some testing and then make changes to here even if the changes don't seem like they'll break anything

src/builtin/welcome.ts
```diff
@@ -9,7 +9,7 @@ const msg_types = Sphinx.MSG_TYPE
```
kevkevinpal, line 9: same as other `src/builtin` files

kevkevinpal:
added some comments the `src/builtin` files and greenlight files don't have test coverage so it's probably best to avoid making changes to them till there are tests that involve them. To avoid any breakage

src/grpc/lightning.ts
antonilol: eslint complains about 'promise executor functions' being async, when removing the async there, await doesn't work anymore inside this function so i moved it a line, also somewhere an rpc was loaded two times inside the function and moving it up made it load only once

src/builtin/welcome.ts
antonilol: what i changed in these files are javascript mechanics, such as adding `{` and `}` around a case statement because eslint complains about a constant being accessible from outside it, or i moved it entirely in 1 return statement to not declare a const

antonilol:
also, all `await` keywords i removed have no effect, my ide tells me the function can not return a promise / is not async. this should not break anything, however i understand that you want to be sure to not break anything

src/grpc/lightning.ts
kevkevinpal: ok cool that is a good improvement then

@Evanfeenstra let me know what you think

kevkevinpal:
going to close this PR for now since it'll need to be rebased but I did include some of these changes in other PR's

Issue #343, opened 2022-04-01
New socket.io CORS


Evanfeenstra:
Getting "Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource" when connecting to new socket.io from win/linux desktop (which uses Electron so CORS is an issue)

@kevkevinpal

kevkevinpal:
ok gotcha is that just an issue on the win/linux desktop? I think I know how to fix this issue but we'll need to test again on the other platforms aswell

kevkevinpal:
Merged Change closing now

PR #344, opened 2022-04-02
using cors instead of preflightRequest attribute for socketio

kevkevinpal:
https://github.com/stakwork/sphinx-relay/issues/343

I think the issue was that the updated socketio doesn't allow cross origin by default so you need to manually say so, need to test this change before merging in

https://socket.io/docs/v3/migrating-from-2-x-to-3-0/#cors-handling

Wonder if there's a way to test this in the socketio test


Evanfeenstra:
From the docs it looks like `origin` should be set to `req.headers.origin`, not `true` ?

kevkevinpal:
> From the docs it looks like `origin` should be set to `req.headers.origin`, not `true` ?

I was reading from here 
https://github.com/expressjs/cors#configuration-options

Where it says to use `true` when we want to use `req.headers("Origin")`

Otherwise I'm not sure where'd we get `req` from unless we can get it from the cors option

Evanfeenstra:
> > From the docs it looks like `origin` should be set to `req.headers.origin`, not `true` ?
> 
> I was reading from here https://github.com/expressjs/cors#configuration-options
> 
> Where it says to use `true` when we want to use `req.headers("Origin")`
> 
> Otherwise I'm not sure where'd we get `req` from unless we can get it from the cors option

Got it. Can we merge this into a branch of `sphinx-relay`... so i can easily just update my live node and test?

kevkevinpal:
> > > From the docs it looks like `origin` should be set to `req.headers.origin`, not `true` ?
> > 
> > 
> > I was reading from here https://github.com/expressjs/cors#configuration-options
> > Where it says to use `true` when we want to use `req.headers("Origin")`
> > Otherwise I'm not sure where'd we get `req` from unless we can get it from the cors option
> 
> Got it. Should be all good then. Can we merge this into a branch of `sphinx-relay`... so i can easily just update my live node and test?

Yea let me push it up and tell you which branch it is one sec

kevkevinpal:
@Evanfeenstra it should be under this branch
https://github.com/stakwork/sphinx-relay/tree/socketIOCrossOrigin

Evanfeenstra:
@kevkevinpal it works like a charm!

Evanfeenstra:
Lets merge this one

PR #352, opened 2022-04-12
small commit to test testing workflow with lnd on stack instead of lnzap/lnd

kevkevinpal:


kevkevinpal:
looks like its working closing now

PR #353, opened 2022-04-13
Upgrading to enable AMP

kevkevinpal:
Added a test that will let us know if amp is working

basically sending 1,500,000 sats from Alice -> Bob when Alice has two 1 mil sat channels open

The only issue I'm having rn is that when sending the payment by running the test we are getting a 
`2022-04-13 15:47:15.270 [ERR] RPCS: [/routerrpc.Router/SendPaymentV2]: destination doesn't support AMP payments`

Which I am trying to debug now

still need to add backwards compatibility support for keysend fallback and additional testing to ensure this works on all types of nodes


Evanfeenstra:
@kevkevinpal the receiving node needs `accept-amp=true` in its LND config

kevkevinpal:
Closing since this is being worked on here

https://github.com/stakwork/sphinx-relay/pull/362

PR #360, opened 2022-04-15
fix kick msg not to all members

Evanfeenstra:


Evanfeenstra:
@kevkevinpal This was changed here https://github.com/stakwork/sphinx-relay/blame/master/src/controllers/chats.ts#L93

the `network.sendMessage` function duplicates the message for each Contact listed in the `contactIds` array on the `chat` param. So for certain ones like this (kicking out a member), only the one contact should receive the message, not the entire tribe

kevkevinpal:
> @kevkevinpal This was changed here https://github.com/stakwork/sphinx-relay/blame/master/src/controllers/chats.ts#L93
> 
> the `network.sendMessage` function duplicates the message for each Contact listed in the `contactIds` array on the `chat` param. So for certain ones like this (kicking out a member), only the one contact should receive the message, not the entire tribe
> 
> There are other controllers that do similar stuff, like some of the "paid attachment" logic. So when working on eslint stuff, make sure to keep the logic consistent!

ya weird not sure why it was commented out, I wonder if running `eslint --fix` does that I'll double check through that PR again to make sure no weirdness happened

kevkevinpal (approved):

PR #363, opened 2022-04-18
direct payments in tribes

Evanfeenstra:
- new fields on Msg: `recipientAlias` and `recipientPic` for direct payments in tribes
- `pay` boolean on sendMessage to trigger a direct payment instead of a simple text msg


kevkevinpal:
@Evanfeenstra you can update the `--max-warnings` in `package.json` to `480` since it seems we've reduced the warning count from 518

PR #370, opened 2022-05-12
update readme

antonilol:
min channel size is 100k sats, lnd keeps 10k sats per anchor channel (100k max) and another 10k sats so the user can allocate it to fee to around 65 sat/vb which should be enough

`--sat_per_byte` is deprecated and replaced by `--sat_per_vbyte`

table looks way nicer now :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,16 +41,16 @@ $ lncli newaddress p2wkh
 }
 ```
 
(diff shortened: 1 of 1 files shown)


kevkevinpal:
looking good thanks for the change :)

LGTM!

PR #362, opened 2022-04-18
AMP payments

antonilol:
just here for the ci tests

PR #373, opened 2022-05-20
automatically update proto files

antonilol:
in #362 i once got an error that a certain field of a rpc response didn't exist when it should have, later i found out these proto files where outdated and the field was missing.

to prevent these types bugs in the future this workflow runs every month, it pulls the proto files from the lnd repository and updates them.

```test -e proto/`basename {}` &&``` ensures only proto files that already exist are updated

greenlight stuff is untouched and rpc.proto had to be renamed to lightning.proto because lnd did

### DIFF
--- a/.github/workflows/update_proto.yml
+++ b/.github/workflows/update_proto.yml
@@ -1,0 +1,26 @@
+name: Update proto files
+on:
+  schedule:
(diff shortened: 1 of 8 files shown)


Evanfeenstra:
this is cool

PR #374, opened 2022-05-23
set fail-fast to false in integration_test.yml 

antonilol:
set fail-fast to false in integration_test.yml to not cancel other tests if one fails

found this in some other workflow and thought this would be handy here too


kevkevinpal:
This looks good, will help identify if multiple tests are failing

PR #376, opened 2022-05-24
testing to see if the proxy integration tests will work

kevkevinpal:
don't mind this PR just checking to see if the tests will still work if we merge the changes to the stack for proxy

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -25,7 +25,7 @@ jobs:
         npm install && npm run build && docker build -t sphinxlightning/sphinx-relay . 
     - name: Checkout stack
       run: |
-        git clone https://github.com/stakwork/sphinx-stack.git stack
+        git clone -b proxy https://github.com/kevkevinpal/sphinx-stack.git stack
     - name: give permissions 
       working-directory: ./stack
       run: |


kevkevinpal:
@Evanfeenstra looks like I figured out how to get the tests to work some are failing rn but lemme see if just running them again will fix it

kevkevinpal:
So one current issue I see with stack with proxy is that sometimes dave.sphinx doesn't get fully setup and is stuck on that is_setup phase so whenever we call that endpoint we get a 404 so the test runs forever

kevkevinpal:
another issue is that the meme server is giving a connection refused when running tests even with the non lite nodes

kevkevinpal:
This is whats causing dave.sphinx to stall the testing env sometimes

Screen Shot 2022-05-27 at 9 10 47 AM

kevkevinpal:
seems like also two tests are legitimately failing for the proxy nodes, the messageLength.test and the images.test not sure if its an issue with the test or if something is broken will need to check into them

kevkevinpal:
Seems like the images test is failing only sometimes on my local machine

seems like the cause is in the sendImage.ts file where we do
`encrypt(n2contactP1.contact_key, upload.media_key)`
the contact_key is coming as null

PR #377, opened 2022-05-25
warnings count to 466

antonilol:
parts of #338 that have test coverage + dev depends from #287

### DIFF
--- a/src/controllers/payment.ts
+++ b/src/controllers/payment.ts
@@ -1,4 +1,4 @@
-import { Message, Chat, models } from '../models'
+import { Message, models } from '../models'
 import { sendNotification } from '../hub'
 import * as socket from '../utils/socket'
 import * as jsonUtils from '../utils/json'
@@ -118,9 +118,9 @@ export const sendPayment = async (req: Req, res: Res): Promise<void> => {
 
   // if contact_ids, replace that in "chat" below
   // if remote text map, put that in
-  let theChat = chat
+  const theChat = chat.dataValues;
   if (contact_ids) {
-    theChat = { ...chat.dataValues, contactIds: contact_ids } as Chat
+    theChat.contactIds = contact_ids
     if (remote_text_map) msgToSend.content = remote_text_map
   }
   network.sendMessage({
--- a/src/grpc/lightning.ts
+++ b/src/grpc/lightning.ts
@@ -171,8 +171,8 @@ export async function queryRoute(
       routes: [],
     }
   }
-  return new Promise(async function (resolve, reject) {
-    const lightning = await loadLightning(true, ownerPubkey) // try proxy
+  const lightning = await loadLightning(true, ownerPubkey) // try proxy
+  return new Promise((resolve, reject) => {
     const options: { [k: string]: any } = { pub_key, amt }
     if (route_hint && route_hint.includes(':')) {
       const arr = route_hint.split(':')
@@ -202,8 +202,8 @@ export type NewAddressType = 0 | 1 | 2 | 3
 export async function newAddress(
   type: NewAddressType = NESTED_PUBKEY_HASH
 ): Promise<string> {
-  return new Promise(async function (resolve, reject) {
-    const lightning = await loadLightning()
+  const lightning = await loadLightning()
+  return new Promise((resolve, reject) => {
     lightning.newAddress({ type }, (err, response) => {
       if (err) {
         reject(err)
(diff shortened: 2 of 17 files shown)


antonilol:
two `@types` packages were causing a lot of errors, removed them for now

antonilol:
running all tests on each `@types` dependency to find where the errors come from https://github.com/antonilol/sphinx-relay/pulls

antonilol:
all tests were successful and adding them here did not give any errors, but previously it did so there must have been some outside things causing this, but anyway, `All checks have passed`

src/controllers/payment.ts
```diff
@@ -120,11 +121,11 @@ export const sendPayment = async (req: Req, res: Res): Promise<void> => {
   // if remote text map, put that in
   let theChat = chat
   if (contact_ids) {
-    theChat = { ...chat.dataValues, contactIds: contact_ids } as Chat
+    theChat = <ChatRecord>{ ...chat.dataValues, contactIds: contact_ids }
```
kevkevinpal, line 124: wondering why are we using ChatRecord and Partial<ChatPlusMembers> instead of just Chat, unless there's something I'm missing

antonilol: `chat` is a `ChatRecord` (returned by `helpers.findOrCreateChat`), then `theChat` is given the same type, but then assigned with `chat.dataValues`. it must be just a `Chat`

antonilol: fixed, and this code also looks better, less hacky type casts

src/grpc/lightning.ts
```diff
@@ -599,7 +587,7 @@ export function signBuffer(msg: Buffer, ownerPubkey?: string): Promise<string> {
         const recidBytes = sigBuf.subarray(66, 67)
         // 31 is the magic EC recid (27+4) for compressed pubkeys
         const ecRecid = Buffer.from(recidBytes).readUIntBE(0, 1) + 31
-        const finalRecid = Buffer.from('00', 'hex')
+        const finalRecid = Buffer.allocUnsafe(1)
```
kevkevinpal, line 590: not sure why we're changing this here does eslint not like the way we did it previously?

when I do this script its giving me false
```
var buf1 = Buffer.from('00', 'hex')
var buf = Buffer.allocUnsafe(1);
console.log(buf1 == buf);
console.log(buf1)
console.log(buf)
```

kevkevinpal: not sure if I fully understand how Buffer works will need to look into it a bit

kevkevinpal:
Other than the one comment and the merge conflicts this PR seems fine to me and I'm good to merge

@Evanfeenstra let me know if you have any comments

src/grpc/lightning.ts
antonilol: its a minor change that runs faster looks cleaner.

your code snippet returns false because of javascript references (the buffers point to different places in memory despite being the same)

differences between buffer initialization methods:

`Buffer.from(string, 'hex')`: buffer with `string.length/2` length and contents of the string
`Buffer.alloc(lenght)`: buffer with `length` length filled with zeros (like c calloc)
`Buffer.allocUnsafe(length)`: buffer with `length` length, just allocated, not touched so can contain garbage (like c malloc)

the first one is the slowest, the last one the fastest.
the buffer is rewritten later in the code so allocUnsafe suits this case best

kevkevinpal: gotcha that makes sense and yea we rewrite over soon after so this looks good 🚀

antonilol:
i rebased and fixed conflicts, but it gave an extra warning, so fixed 4 simple ones in utils/gitinfo.ts

kevkevinpal:
seems like a new a new error in the tests
```
src/utils/cert.ts:208:9 - error TS2322: Type 'KeyPair' is not assignable to type '{ [k: string]: string; }'.
  Index signature for type 'string' is missing in type 'KeyPair'.

208   const transportTokenKeys: { [k: string]: string } = await rsa.genKeys()
```

antonilol:
> seems like a new a new error in the tests
> 
> ```
> src/utils/cert.ts:208:9 - error TS2322: Type 'KeyPair' is not assignable to type '{ [k: string]: string; }'.
>   Index signature for type 'string' is missing in type 'KeyPair'.
> 
> 208   const transportTokenKeys: { [k: string]: string } = await rsa.genKeys()
> ```

oh oops that was 'the missing warning', probably some code got moved, fixed now

kevkevinpal:
LGTM 🚀 

@Evanfeenstra I'm merging now but lemme know if you see anything that you'd like to change

Issue #379, opened 2022-05-26
Feature request: Core Lightning support


AaronDewes:
Core Lightning would be great to have in Sphinx so it's not limited to LND users. While you already support greenlight, self-hosted Core Lightning would allow more users to use Sphinx.

andrewlunde:
Plus 1 from me.

Evanfeenstra:
GRPC is ready enough now to do this

https://github.com/ElementsProject/lightning/blob/master/cln-grpc/proto/node.proto

AaronDewes:
> GRPC is ready enough now to do this
> 
> [https://github.com/ElementsProject/lightning/blob/master/cln-grpc/proto/node.proto](https://github.com/ElementsProject/lightning/blob/master/cln-grpc/proto/node.proto?rgh-link-date=2022-06-27T18%3A15%3A56Z)

I've also built a TS client for it: https://github.com/runcitadel/core-ln.ts (It uses the JSON-RPC API though, not the gRPC API)

Evanfeenstra:
started here: https://github.com/stakwork/sphinx-relay/tree/feat/cln

MASTERCOSMOS:
That should be fine.

PR #383, opened 2022-06-08
grpc typescript types

antonilol:
after a long search to proto file to typescript types compiler i found that `@grpc/proto-loader` had one (in a newer version), the only one i got it working with. the types are correct, camel case is preserved

there are also some things in the codebase that are hard to work with, like greenlight `Node`, proxy `Lightning` and lnrpc `Lightning` are all in one var, and where an rpc is used there is an `IS_GREENLIGHT` check, or the `...Command` functions that return the specific rpc command for lnd and greenlight (`lightning[cmd]()`)

i also saw an issue posted recently to add cln support (#379), just adding will make the above issue even worse, so i like to make a 'layered' abstract lightning api that has all lightning interactions used by sphinx in a set format (ts interfaces), and on top of these abstract functions are the backend specific functions, this way have one input and output type, no matter the backend.

(btw im not a cln user myself (yet), have to figure out first how calls work there)


Evanfeenstra:
ya this is what i was trying to do in this file @antonilol https://github.com/stakwork/sphinx-relay/blob/master/src/grpc/interfaces.ts

making it so greenlight GRPC (or CLN next) can be abstracted away, and dealt with just like its LND

antonilol:
> ya this is what i was trying to do in this file @antonilol https://github.com/stakwork/sphinx-relay/blob/master/src/grpc/interfaces.ts
> 
> making it so greenlight GRPC (or CLN next) can be abstracted away, and dealt with just like its LND

oh yeah, i see, but having multiple implementation's apis in one type can make debugging harder (the goal of typescript is catching most of em but with types `any` this wont work), and if in the future a lightning implementation that does not have grpc is added this part of the code will become a nested `if` hell
also, lnd's api responses can also change, will every other response be changed then as well to match lnd's response? i prefer all api do their implementation specific things and returning it in a format sphinx can use and gets all needed information

antonilol:
will be replacing this with a new pr that also removes the deprecated `grpc` in favor of `@grpc/grpc-js`, testing this now and running tests on my fork. will be opening the pr once the package lock problem is fixed and i rebased on top of it (i requires dependency changes)

kevkevinpal:
> will be replacing this with a new pr that also removes the deprecated `grpc` in favor of `@grpc/grpc-js`, testing this now and running tests on my fork. will be opening the pr once the package lock problem is fixed and i rebased on top of it (i requires dependency changes)

have this PR up will merge once tests start passing

https://github.com/stakwork/sphinx-relay/pull/422

PR #387, opened 2022-06-13
Revert "Revert "update some npm dependencies because some had critical vulnerabilities""

antonilol:
Reverts stakwork/sphinx-relay#384


antonilol:
now with the new nodejs version

kevkevinpal:
> @kevkevinpal these tests are all successful (except 1 is still running)

yup looking good :) seems like the socket.io test fails randomly, it did the same on the yml file PR change too but I just re ran it

kevkevinpal:
LGTM 🚀

PR #390, opened 2022-06-15
Added long message test

DitchRat:
created change to solve this issue: https://github.com/stakwork/sphinx-stack/issues/18

pretty much test just creates very long messages and sends message from node 1 to node 2 and confirms to make sure the 2 messages are the same.


kevkevinpal:
Looks like the test is passing @DitchRat 

@Evanfeenstra can you double check to see if this test looks good/ if the message is long enough, then we can probably merge it

kevkevinpal:
merging changes to test

Evanfeenstra:
looks good!

PR #393, opened 2022-06-27
Fix/gitbot events

Evanfeenstra:
better event parsing for gitbot webhooks


src/utils/githook.ts
```diff
@@ -154,21 +184,31 @@ export function processGithook(
...
       }
     }
   }
   return ''
 }
+
+function trunc(str: string) {
+  return truncateString(str, MAX_MSG_LENGTH)
+}
+function truncateString(str: string, num: number) {
```
kevkevinpal, line 209: are we able to have one function with an optional param set as `MAX_MSG_LENGHT` instead of two functions?

src/utils/githook.ts
```diff
@@ -118,33 +138,43 @@ function pushAction(e: PushEvent): string {
...
 }
-const actionsMap: { [k in PropName]: ActionMap } = {
-  issue: issueActionMap,
+function deleteAction(e: DeleteEvent): string {
+  if (e.ref_type === 'branch') {
+    return `Branch ${e.ref} deleted in ${e.repository.full_name}`
+  } else if (e.ref_type === 'tag') {
+    return `Tag deleted in ${e.repository.full_name}: ${e.ref}`
+  } else {
+    return ''
```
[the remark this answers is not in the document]
Evanfeenstra: empty string is skipped

src/utils/githook.ts
Evanfeenstra: i think its fine like this :) because possibly there will be another different truncate with a different length at some point later on?

PR #397, opened 2022-07-06
[TEST] New test for unique aliases

divyank-rs:
added new test for checking that logic for assigning unique alias in case of duplicate aliases in tribe is working


kevkevinpal:
Looking good will try and merge it today just waiting on this final test to pass, Also ticket may remain up for a bit on the sphinx.people page since my node is taking a bit to connect to the tor network but feel free to move onto the next ticket and I can also disperse the bounty

PR #407, opened 2022-08-04
proxy lsat table migration from STRING to TEXT values

kevkevinpal:
Was getting an error of `value too long for type character varying(255)` when trying to call the `saveLsat` endpoint but I made some changes to `src/utils/migrate.ts` to change the type to a larger value of type `TEXT` since `STRING` seemed to be giving the error


Evanfeenstra:
dont think `alter table x alter x type x` is supported in sqlite

kevkevinpal:
closing cause another PR opened to fix this issue

PR #408, opened 2022-08-05
added an updated sql file to use the TEXT feild instead of string sin…

kevkevinpal:
…ce we get an error for longer texts like lsats and macaroons


Evanfeenstra:
👍

PR #410, opened 2022-08-15
Update to node16

kevkevinpal:
Updating sqlite to version 5 from 4 so we may need to double check things are working here

But seems like with the update sequilize is returning an unknown type so we need to type a lot of these database calls

Draft for now just updating the types for each file, I think prettier might be also applying in the diff but its only been type changes


src/notify.ts
```diff
@@ -101,7 +102,9 @@ const sendNotification = async (
     try {
       const cids = JSON.parse(chat.contactIds || '[]')
       const notme = cids.find((id) => id !== 1)
-      const other = models.Contact.findOne({ where: { id: notme } })
+      const other: Contact = models.Contact.findOne({
```
kevkevinpal, line 105: is this correct should we not await this command? cause also other.blocked wouldn't be an attribute on a promise would it?

kevkevinpal:
I finished doing all the files that were relatively straight forward just have a few more which are a bit more interesting but I will do them in their own commits

Evanfeenstra:
looks like 2 tests failed?

kevkevinpal:
> looks like 2 tests failed?

ya seems like just outlier failures should pass if I rerun, still need to make some changes for the lint test but almost there

src/controllers/chatTribes.ts
```diff
@@ -460,7 +480,7 @@ export async function approveOrRejectMember(req: Req, res) {
 
   network.sendMessage({
     // send to the requester
-    chat: { ...chatToSend, contactIds: [member.contactId] },
+    chat: { ...chatToSend, contactIds: member.contactId.toString() },
```
kevkevinpal, line 483: This was giving a weird type so I changed this, would this break anything?

src/utils/timers.ts
```diff
@@ -75,16 +76,21 @@ export async function reloadTimers() {
...
+    where: { id: t.tenant },
+  })) as ContactRecord
   if (!chat) {
     models.Timer.destroy({ where: { id: t.id } })
     return
   }
-  const theChat = { ...chat.dataValues, contactIds: [t.receiver] }
+  const theChat: Partial<ChatPlusMembers> = {
+    ...chat.dataValues,
+    contactIds: t.receiver.toString(),
```
kevkevinpal, line 92: This is definitely breaking a test but the types don't match

kevkevinpal:
These last 8 commits with `[*]` have the potential to break something the rest are regular type changes

kevkevinpal:
I see this bit of code that might have been affected, seems like its accounting for the type discrepancy. Will need to see how to do this more elegantly and work with the changes

https://github.com/stakwork/sphinx-relay/blob/master/src/network/send.ts#L73

kevkevinpal:
@Evanfeenstra the tests are working now and should allow us to build to node version 16

I would double-check any of the commits with `[*]` at the beginning they're the changes that are not super straightforward

Evanfeenstra (approved):

kevkevinpal:
few updates after testing on my node

see the following two in the logs but relay seems to be working fine for me

---
Screen Shot 2022-08-22 at 9 01 29 PM

----

Screen Shot 2022-08-22 at 8 52 36 PM

kevkevinpal:
Merging the node16 changes

PR #411, opened 2022-08-21
Unit tests (Jest)

kevkevinpal:
adding a new type of test other than the integration test we have

added a simple test case for the toSnake and toCamel utils file to show how it'll work

I'm pretty sure we can add enforcement like % code coverage and etc but we can start with just having the tests for now

Added a gitaction too to show us if we break any testing


kevkevinpal:
Merging these changes more unit tests to come

PR #412, opened 2022-08-24
Feat/push

Evanfeenstra:


kevkevinpal:
This is the error I'm seeing when I'm running the boostMessage test

Screen Shot 2022-08-25 at 11 41 28 AM

kevkevinpal:
I think in detect message when the message has no content it throws an error, in this example its a boost

@Evanfeenstra maybe check if it's undefined and then handle it?

kevkevinpal:
Spotted an error when running this branch
Screen Shot 2022-09-01 at 11 25 33 AM

Evanfeenstra:
> Spotted an error when running this branch 

Just pushed a fix

Evanfeenstra:
woooohooo finally all the tests pass!

kevkevinpal:
Tried doing an @ and got this error

Screen Shot 2022-09-02 at 10 17 39 AM

kevkevinpal:
So from what I'm seeing it looks like the admin is not in the `sphinx_chat_members` table and that you can have a user with a space in their name which kinda breaks how the parsing of the @ is implemented now

Evanfeenstra:
Pushed a fix for that issue. But ya we need to think about the tribe admin being mentioned as well...

PR #419, opened 2022-09-07
Feat/knowledge graph

tobi-bams:


kevkevinpal:
this looks good to me, but maybe we can keep it on a separate branch for now instead of trying to merge to master. So we can do some testing first

But otherwise it looks great

kevkevinpal:
closing this PR since we'll be using https://github.com/stakwork/sphinx-relay/pull/424

Issue #420, opened 2022-09-07
CVE-2022-24434


antonilol:
noticed this output recently when looking at the logs of the integration tests. i don't get this message when running `npm i` on my local machine on a freshly cloned relay...
```
npm WARN deprecated multer@1.4.4: Multer 1.x is affected by CVE-2022-24434. This is fixed in v1.4.4-lts.1 which drops support for versions of Node.js before 6. Please upgrade to at least Node.js 6 and version 1.4.4-lts.1 of Multer. If you need support for older versions of Node.js, we are open to accepting patches that would fix the CVE on the main 1.x release line, whilst maintaining compatibility with Node.js 0.10.
```

node -v: v16.16.0, npm -v: 8.18.0

kevkevinpal:
which integration test do you see it on?

kevkevinpal:
ok I see it 
https://github.com/stakwork/sphinx-relay/runs/8226373631?check_suite_focus=true#step:6:786

looks like its on the docker `RUN npm install bcrypt` step

kevkevinpal:
yea not sure it was there before I started working on this project @Evanfeenstra do you know why we have it there

I see its added in this commit
https://github.com/stakwork/sphinx-relay/commit/19492181d3b131aed9efe857da10f5fbaffd330f#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

kevkevinpal:
Looks like all the integration tests passed when I removed bcrypt

PR #421, opened 2022-09-07
removed bcrypt

kevkevinpal:
checking to see if the integration tests will fail if we remove bcrypt

if it works it would close https://github.com/stakwork/sphinx-relay/issues/420

probably more checking may need to be done tho


kevkevinpal:
Looks like its working without bcrypt merging

PR #423, opened 2022-09-08
switch out `grpc` for `@grpc/grpc-js` + ts types

### LINKED ISSUE #295: build fails on node.js v17 due to depreciated grpc
This is the upstream bug that is the culprit.  The proposed solution is to drop grpc for @grpc/grpc-js
https://github.com/grpc/grpc-node/issues/2001

### LINKED ISSUE #425: @grpc/grpc-js $@500000
switch to `@grpc/grpc-js`... but keep it camel_case

antonilol:
fixes #295
fixes #425

i managed to get `@grpc/grpc-js` working with camel_case fields, ~but still some tests are failing~

also added generated typescript types for the grpc calls

all files in src/grpc/types/ are generated with the `./grpc_gen_types.sh` bash script, which also runs in gh workflows after updating the proto files


antonilol:
what does `failed to getContactAndCheckKeyExchange` actually mean? i see it in most of the failed tests. a more descriptive error would help

Evanfeenstra:
> 

When a contact is added, you send a keysend with your `contact_key` (end-to-end encryption key)... then they send a keysend back with their `contact_key`

antonilol:
> When a contact is added, you send a keysend with your `contact_key` (end-to-end encryption key)... then they send a keysend back with their `contact_key`

ok got it, i think it didnt got a key back from the other party then...

Evanfeenstra:
@antonilol https://github.com/stakwork/sphinx-relay/issues/425

antonilol:
@Evanfeenstra  can you fix package-lock.json on master? it contains merge conflict markers...

Evanfeenstra:
> @Evanfeenstra can you fix package-lock.json on master? it contains merge conflict markers...

ok @antonilol its fixed

Evanfeenstra:
https://github.com/stakwork/sphinx-relay/pull/423/files#diff-acc59c19f30ca8e2dd83cb49828c32894fce27500a1fb8bdc6320d148cb4257bL68

Does that need to be `return await loadProxyLightning(ownerPubkey)` @antonilol ?

antonilol:
lets try

Evanfeenstra:
> lets try

that will not fix the errors tho, the stack tests are not build around proxy. I would recommend cloning `sphinx-stack` and building `sphinx-relay` locally, then using your local image in sphinx-stack. Then you can see all the logs and what is going on (with Docker Desktop is easiest) @antonilol

antonilol:
> > lets try
> 
> that will not fix the errors tho, the stack tests are not build around proxy. I would recommend cloning `sphinx-stack` and building `sphinx-relay` locally, then using your local image in sphinx-stack. Then you can see all the logs and what is going on (with Docker Desktop is easiest) @antonilol

yeah i wanted to do this but stack comes with all types of random errors, like the one i described here https://github.com/stakwork/sphinx-stack/issues/45 (i managed to fix this by running `docker-compose up -d` en then `docker-compose down`, and after that the no alice config seemed to work)
now i am getting an endless spam of `=> try  http://host.docker.internal:3001/is_setup` in `docker logs -f sphinx-stack-relaysetup-1` (i believe port 3001 is from alice.sphinx but there is no alice...)
and bob and carol report this:
```
(node:1) DeprecationWarning: grpc.load: Use the @grpc/proto-loader module with grpc.loadPackageDefinition instead
[info] 09-15-22T18:49:53 [MISC] Node listening on 3002.
=> initGrpcSubscriptions error Error: 12 UNIMPLEMENTED: unknown service lnrpc.Lightning
    at Object.exports.createStatusError (/relay/node_modules/grpc/src/common.js:91:15)
    at Object.onReceiveStatus (/relay/node_modules/grpc/src/client_interceptors.js:1209:28)
    at InterceptingListener._callNext (/relay/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/relay/node_modules/grpc/src/client_interceptors.js:618:8)
    at callback (/relay/node_modules/grpc/src/client_interceptors.js:847:24) {
  code: 12,
  metadata: Metadata { _internal_repr: {}, flags: 0 },
  details: 'unknown service lnrpc.Lightning'
}
```

antonilol:
i got stack working by running just `docker-compose up -d`, waiting till everything is running and then stopping alice.sphinx, then running sphinx relay and at least i got one test to pass on my machine (queryroutes)

antonilol:
i got some more tests working now with only the bare minimum to get grpcjs working, without some suggested type checks

antonilol:
i got all tests passing locally!!!
i squashed all commits and now waiting for the gh ci tests to pass

Evanfeenstra:
> i got all tests passing locally!!! i squashed all commits and now waiting for the gh ci tests to pass

what was the issue that was stopping key exchange?

antonilol:
> > i got all tests passing locally!!! i squashed all commits and now waiting for the gh ci tests to pass
> 
> what was the issue that was stopping key exchange?

there were 2 issues:
1. keysend didnt work because grpcjs thought a bytebuffer always had a length of 0, i replaced it with a nodejs native buffer
2. `map<uint64, something>` in proto will cause grpcjs to hash the uint64 before giving it to the client, resulting in the keys of the dest_custom_records being hashed, then no key `133773310` existed anymore

Evanfeenstra:
@antonilol so this file https://github.com/stakwork/sphinx-relay/blob/97ea3a3a34e42d5102c9396760ba706a0206d176/protobuf_long.patch is needed in order to make the `map<uint64, something>` work? Do we need to do anything specific to make it work? I'm not familiar with npm patches

antonilol:
> @antonilol so this file https://github.com/stakwork/sphinx-relay/blob/97ea3a3a34e42d5102c9396760ba706a0206d176/protobuf_long.patch is needed in order to make the `map<uint64, something>` work? Do we need to do anything specific to make it work? I'm not familiar with npm patches

Yes
I added a `postinstall` script to apply the patch to the package.json, this gets automatically ran by npm after `npm install`

kevkevinpal:
Hey @antonilol were you able to manually test this change I pulled in the changes and I was only able to send out messages, receive wasn't working

do you know if this would fail if you're running nodejs 12 or is it required it be updated?

antonilol:
> Hey @antonilol were you able to manually test this change I pulled in the changes and I was only able to send out messages, receive wasn't working
> 
> do you know if this would fail if you're running nodejs 12 or is it required it be updated?

i use node v16, and it works. with node v12.22.12 i cant npm install because this protobuf patch doesnt apply. if i just start it on node v12 i get an error that mcrypt was compiled for a newer version of nodejs

antonilol:
rebased

kevkevinpal:
Ok was able to get this running on nodejs 16 and it seems to be working fine with manual testing, just need to remember to update things from node 12 to 16 now

kevkevinpal (approved):
Looks like linter is failing but otherwise I am good with these changes

kevkevinpal:
> > Looks like linter is failing
> 
> this is fixed in #445, eslint misses some files and previous changes added warnings, can you merge it before this one?

merged

antonilol:
rebased, got the warning count right and lint check passed now

antonilol:
all tests passed except one port in use error (https://github.com/stakwork/sphinx-stack/issues/47)

kevkevinpal:
> all tests passed except one port in use error ([stakwork/sphinx-stack#47](https://github.com/stakwork/sphinx-stack/issues/47))

test are passing now 🥳

antonilol:
rebased and ready to merge. did you manage to upgrade the lite nodes to nodejs v16?

kevkevinpal:
> rebased and ready to merge. did you manage to upgrade the lite nodes to nodejs v16?

Sorry seeing this now, I'm trying to upgrade the nodes on our end just had to add more storage since the clusters we're getting full after trying to reinstall the node_modules. I have a script to auto update it just might take a long time update, I'm getting stuck on this node_modules when installing `[##################] / reify:macaroon: timing reifyNode:node_modules/typescript` not sure if you know why this in one in particular takes so long

antonilol:
> I'm getting stuck on this node_modules when installing `[##################] / reify:macaroon: timing reifyNode:node_modules/typescript` not sure if you know why this in one in particular takes so long

On my devices it also takes more time than other things done during `npm install`, on my bitcoin node it was significantly slower than on my pc and laptop, i think it is some cpu bound task, like compiling

antonilol:
rebased this and #362

kevkevinpal:
> rebased this and #362

thanks for rebasing I'm still in the process updating the clusters just ran into some issues and it takes a while to update each one so I'm doing them at night

Thanks for your patience

PR #435, opened 2022-09-16
run prettier in github actions

antonilol:
title

~draft pr for now because i dont want this merged before #423 (it will be a pain to rebase that pr with many little changes this produces)~


kevkevinpal:
is there anyway of running prettier on the PR itself when someone opens to master so the diff is never ugly incase someone auto formats to prettier with their own config?

antonilol:
> is there anyway of running prettier on the PR itself when someone opens to master so the diff is never ugly incase someone auto formats to prettier with their own config?

I tried this with the auto build but it didnt seem to work, even with the "Allow edits from maintainers" setting turned on

antonilol:
just realized i can run prettier on other prs and merged should go fine because both branches contain the same change

antonilol:
rebased + also format `.github/**/*.yml` now

antonilol:
rebased

kevkevinpal (approved):
This looks good to me

PR #439, opened 2022-09-26
Feat/get update lsat

tobi-bams:
@kevkevinpal 
Here is the get active and update last endpoint


kevkevinpal (approved):
looks good to me @Evanfeenstra lemme know what you think

PR #443, opened 2022-10-04
[FEAT] added a route for admin to be able to delete sphinx tickets

divyank-rs:


Evanfeenstra:
@divyank-rs you do not need to run `const owner: Contact = await models.Contact.findOne`. That query is run automatically on all authenticated requests, and the owner is available at `req.owner`

PR #444, opened 2022-10-06
Feat/settle taggers

tobi-bams:
Settle taggers by boost and stream


kevkevinpal:
from what we talked about it might make more sense to put it on `sphinx_action_history` instead of a new table

PR #445, opened 2022-10-09
fix eslint command glob pattern

antonilol:
the pattern `./src/*.ts ./src/**/*.ts` only matches *.ts in src and 1 directory deeper (because npm script dont have bash globstar enabled)
apparently you can give eslint a directory and it takes all files from it, recursively, so 2 files that were more that 1 directory deep in src have now also been auto-fixed and are included in the warning count


antonilol:
btw, while busy with eslint i thought about [this](https://github.com/stakwork/sphinx-relay/issues/334) bounty. can i still get it for pr #377? (51 warns)

antonilol:
just rebased this pr

PR #466, opened 2022-11-04
Update/action history

tobi-bams:
Added a new column to the `action_history` table that helps store action types as integers


kevkevinpal (approved):
ya this looks good to me

kevkevinpal:
we might wanna start to think about adding testing for this stuff, merging for now

PR #467, opened 2022-11-04
Feat/leadership board

tobi-bams:


Evanfeenstra (changes requested):
Very nice. Please update the error message so they are different in the 2 different places. In case there is an error, its best to know exactly where it comes from

src/network/receive.ts
```diff
@@ -235,7 +236,24 @@ async function onReceive(payload: Payload, dest: string) {
...
+            },
+          })) as ChatMemberRecord
+          await sender.update({
+            totalSpent: sender.totalSpent + payload.message.amount,
+          })
+        } catch (error) {
+          sphinxLogger.error(
+            `=> Could not update ChatMember table for Leadership board`,
+            error
+          )
```
kevkevinpal, line 255: sphinxLogger will get confused by that formatting

try something like this
```
sphinxLogger.error(`=> Could not update ChatMember table for Leadership board ${error}`, logging.Network)
```

src/network/receive.ts
```diff
@@ -244,7 +262,7 @@ async function onReceive(payload: Payload, dest: string) {
           owner,
           forwardedFromContactId
         )
-      else sphinxLogger.error(`=> insufficient payment for this action`)
+      } else sphinxLogger.error(`=> insufficient payment for this action`)
```
kevkevinpal, line 265: lets try adding network to the end of this since its in a network folder
```sphinxLogger.error(`=> insufficient payment for this action`, logging.network)```

File: src/network/send.ts [hunk 3]
```diff
@@ -204,6 +205,23 @@ export async function sendMessage({
...
+          },
+        })) as ChatMemberRecord
+        await receiver?.update({
+          totalEarned: receiver.totalEarned + amount,
+        })
+      } catch (error) {
+        sphinxLogger.error(
+          `=> Could not update ChatMember table for Leadership board`,
+          error
+        )
```
kevkevinpal, line 223: Similar to above
sphinxLogger will get confused by that formatting

try something like this
```
sphinxLogger.error(`=> Could not update ChatMember table for Leadership board ${error}`, logging.Network)
```

kevkevinpal (commented):
added comments on sphinxLogger, otherwise looking good

src/tests/configs/nodes.json
```diff
@@ -2,37 +2,37 @@
   {
```
kevkevinpal, line 2: it might be worth adding this to the gitignore since it just makes the PR's look ugly if someone accidentally commits it and its already in the repo, not sure if that would break anything
@Evanfeenstra what do you think?

Evanfeenstra: Definitely lets gitignore it. Im going to merge this for now tho

PR #468, opened 2022-11-05
fix workflow yml auto format

antonilol:
fixes github actions not committing some changes because i forgot to `git add` those files in #435

i can see in the logs that it did format them


antonilol:
```
! [remote rejected] master -> master (refusing to allow a GitHub App to create or update workflow `.github/workflows/integration_test.yml` without `workflows` permission)
```
seems that github doesn't allow editing workflow files from workflows by default, [but this can be changed](https://stackoverflow.com/a/66644086/13800918)

PR #480, opened 2022-11-09
Feat/reputation

tobi-bams:


Evanfeenstra (changes requested):
- in receive, you need `tenant` in the models.Contact.findOne
- dont need "const currentTribe = (await models.Chat.findOn", its called `chat`
- in send, put the leaderboard/reputation stuff in another function and just call it. just to keep it clean

PR #491, opened 2022-11-17
test: getChats to see tribe info

kevkevinpal:
right now the second node is not seeing the pinned messages so I want to add a test to see the new pinned messages

just logging from the getChats call and I'm seeing `PIN` in node1 but not node2

Note: this is not finished yet just adding the tests and then will move onto the implementation, just adding the draft for visibility


kevkevinpal:
I will need to change this test to fetch from the tribes server to check if node2 can get the pinned msg

kevkevinpal:
closing since we already call the tribes server in the test

PR #493, opened 2022-11-18
Feat/badge bot

tobi-bams:


src/controllers/api/personal.ts
```diff
@@ -7,6 +7,9 @@ import * as jsonUtils from '../../utils/json'
 import { success, failure } from '../../utils/res'
 import { loadConfig } from '../../utils/config'
 import { createJWT, scopes } from '../../utils/jwt'
+import { Badge } from '../../types'
+import { createOrEditBadgeBot } from '../../builtin/badge'
```
kevkevinpal, line 11: I'm not sure this makes sense the bot shouldn't be imported elsewhere in relay as relay shouldn't be dependent on the bot the bot should be dependent on relay

tobi-bams: I actually do not get what you mean.....

PR #495, opened 2022-11-21
Feat/person data

tobi-bams:


kevkevinpal:
@tobi-bams Can you add a test to this PR getting the person data and validating that it is correct?

PR #502, opened 2022-11-29
ba-Feat/badge bot test

tobi-bams:


kevkevinpal:
maybe adding functionality to use `/badge create ...badge_args` would be cool so we don't need to wait for a frontend to be built

tobi-bams:
Screenshot 2022-12-06 at 16 46 17

A screenshot of the badge bot working

kevkevinpal:
is there a way to add this info to that message 

`{"id":185,"icon":"earn-badge","name":"Spending","asset":"c6bd66770303542b04044f66d69772d8c72a11dfbe8d2e065893b4e3006f0be5","token":"bb8ef70ea0ec8dc713426c0d2b49cd50c4163fb67dfa5abf135f8ff18c261dbf","amount":10,"creator":"037820f6377ca9df93f15bad6968e5b06f9a87f1bc40c29e2ebeaeab06e0b5df62"},`

or maybe a link to the blockstream liquid explorer with the asset id
https://blockstream.info/liquid/asset/c6bd66770303542b04044f66d69772d8c72a11dfbe8d2e065893b4e3006f0be5

and maybe add self custody the token by heading to the people page

So maybe
```
Tobi earned the spending badge!
https://blockstream.info/liquid/asset/c6bd66770303542b04044f66d69772d8c72a11dfbe8d2e065893b4e3006f0be5
redeem on people.sphinx.chat
```

We can do this for now so people can know exactly where to see the asset

kevkevinpal:
you can find all the assets on the elements server by this endpoint
https://liquid.sphinx.chat/list

tobi-bams:
Alright, this is possible would update that.....

tobi-bams:
@kevkevinpal updated the message for the badge bot response, attached is a screenshot
Screenshot 2022-12-08 at 13 16 06

kevkevinpal:
@tobi-bams to test add `-b add_bolt` here to get the test running using the correct branch and if it works we should be good to merge the sphinx-stack changes and the change it back to master and then merge this change

https://github.com/stakwork/sphinx-relay/pull/502/files#diff-2289f323d1e30f0ec6781fe565465798b0be9d8facf72fad2bd5701e0357a885R64

tobi-bams:
@kevkevinpal yeah, we could use that branch, but for now, the test would still fail because the liquid server is not allowing us to create badges.
We also need to update the boltwall image in Docker Hub

kevkevinpal:
@tobi-bams do you know if the boltwall container is calling the actual sphinx-liquid server?

tobi-bams:
> @tobi-bams do you know if the boltwall container is calling the actual sphinx-liquid server?

Yes it is calling the actual sphinx-liquid server

kevkevinpal:
> > @tobi-bams do you know if the boltwall container is calling the actual sphinx-liquid server?
> 
> Yes it is calling the actual sphinx-liquid server

Hmm if that's the case then it doesn't make sense to do that in the testing env because then we'll run out of sats very quickly since we're running these test for every PR

it seems like either we need to get a mock response somehow coming from the sphinx-liquid server or get the liquid server working on regtest on sphinx-stack which might be difficult considering it needs the liquid sidechain

tobi-bams:
> > > @tobi-bams do you know if the boltwall container is calling the actual sphinx-liquid server?
> > 
> > 
> > Yes it is calling the actual sphinx-liquid server
> 
> Hmm if that's the case then it doesn't make sense to do that in the testing env because then we'll run out of sats very quickly since we're running these test for every PR
> 
> it seems like either we need to get a mock response somehow coming from the sphinx-liquid server or get the liquid server working on regtest on sphinx-stack which might be difficult considering it needs the liquid sidechain

Yea I agree, I would suggest we go with a mock response

kevkevinpal:
I'm thinking if it might take a while to include mock responses to sphinx-stack we might want to do extensive unit testing to get this feature going quicker

you can find the unit tests here

https://github.com/stakwork/sphinx-relay/tree/master/src/tests/unit_tests

Evanfeenstra:
Nice job creating tests for it. What is you plan to add mocks?

I think you need some more validation in places. Like after `const amount = Number(arr[3])` you can check `if (!amount)` then return a failure text message like "Invalid amount"

Also a good idea to include a `/badge help` command that returns some help text (like what order the arguments are in)

tobi-bams:
Thanks @Evanfeenstra, 
I would be using [Mockable.io](https://www.mockable.io/) to mock responses from the liquid server and add this endpoint to the boltwall environment variable.

I would add those extra validations thanks.

In regards to the `/badge help` I have that already

PR #503, opened 2022-11-30
feat: update lsats that are valid before session_secret was set to invalid

tobi-bams:


kevkevinpal (approved):
Looking good to me LGTM 🚀

PR #505, opened 2022-12-03
feat: added logs with chatgpt

kevkevinpal:
Added logs to cert util file using chat gpt

@Evanfeenstra what do you think about this?

I still haven't read over the results too much but they look decent


Evanfeenstra:
> Added logs to cert util file using chat gpt
> 
> @Evanfeenstra what do you think about this?
> 
> I still haven't read over the results too much but they look decent

Holy shit they are pretty damn good

Issue #513, opened 2022-12-15
Chat seen state not updated


tomastiminskas:
When receiving a bot response message on a tribe you are admin of, the `chat.seen` attribute is not being updated to `false` and it should. 

On this line of the `receiveBotRes` method:

https://github.com/stakwork/sphinx-relay/blob/862e43ae6f479d2c8bf1811a9f5d863086fd736d/src/controllers/bots.ts#L456

we should do the same we do on `onReceive` on this line:

https://github.com/stakwork/sphinx-relay/blob/00b9794e1d7187380ca34ce235e9a7fafa39aba8/src/network/receive.ts#L115

kevkevinpal:
so `chat.update({seen: false})` in an if block if there is no chat? @tomastiminskas

tomastiminskas:
` if (chat) chat.update({ seen: false })`

That is the line that should be added before line 456. So if a chat exists, then we update it to seen: false. Let me know if it's clear

PR #514, opened 2022-12-15
fix: made the bot message seen as false

kevkevinpal:
closes https://github.com/stakwork/sphinx-relay/issues/513

@tomastiminskas


tomastiminskas:
We were testing with @gonzaloaune and it is not there where the bot response messages is created. It's actually on this line:

https://github.com/stakwork/sphinx-relay/blob/892d093fe46396de37417df63ab295963d34f4d5/src/controllers/botapi/broadcast.ts#L55

The `chat` in that method is not a chat object, so we need to first use `findOne` to find the Chat Object using chat.id, then update to `seen : false`

@kevkevinpal can you do the change and deploy again to my nodes? I would leave the change you already did because on bots.ts a new message with type bot_res is created and every time that happens the chat object should become unseen.

PR #518, opened 2022-12-19
Debug recommendation

tobi-bams:


kevkevinpal (commented):
Those will be really large logs in relay, I would suggest to only do that for debuging purposes

tobi-bams:
> Those will be really large logs in relay, I would suggest to only do that for debuging purposes

Ops!!!....

So sorry, I am actually using this branch for debugging at the moment. I would remove the logs now

kevkevinpal (commented):
Looks good now 🚀

PR #520, opened 2022-12-22
fix: lower limit amount for /feeds

kevkevinpal:
This is because when we were pulling the action history there is a large size but express is throwing an error when the request size is too large. So I added a todo to compress the size of the request so we can fit more data into it

### DIFF
--- a/src/controllers/getFeeds.ts
+++ b/src/controllers/getFeeds.ts
@@ -11,9 +11,11 @@ export async function getFeeds(req: Req, res: Response) {
   const tenant: number = req.owner.id
   const config = loadConfig()
   try {
+    //TODO: compress the request
+    // limit set to 90 because request size was getting to large
     const actions = (await models.ActionHistory.findAll({
       where: { tenant },
-      limit: 500,
+      limit: 90,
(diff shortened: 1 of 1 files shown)


tobi-bams (commented):
yeah, this works, but did you run the build script???

kevkevinpal:
> yeah, this works, but did you run the build script???

@tobi-bams there is a auto build script in gitactions that @antonilol made
It keeps the diff of the files changed simpler which is why I don't usually add the dist files
https://github.com/stakwork/sphinx-relay/blob/master/.github/workflows/lint_format_and_build.yml

tobi-bams:
> > yeah, this works, but did you run the build script???
> 
> @tobi-bams there is a auto build script in gitactions that @antonilol made It keeps the diff of the files changed simpler which is why I don't usually add the dist files https://github.com/stakwork/sphinx-relay/blob/master/.github/workflows/lint_format_and_build.yml

Alright, that's perfect then...

Let's merge this 🚀

kevkevinpal:
will merge one automated tests are passing

PR #523, opened 2022-12-28
Eslint/tribes promise async

kevkevinpal:


src/utils/tribes.ts
```diff
@@ -56,68 +56,66 @@ async function initializeClient(
   host: string,
   onMessage?: (topic: string, message: Buffer) => void
 ): Promise<mqtt.Client> {
-  return new Promise(async (resolve) => {
```
kevkevinpal, line 59: diff might be hard to read but the main change is getting rid of this promise + async and then just returning the value we get

PR #526, opened 2022-12-29
fix: remvoed deprecated auth code that was not used in codebase

kevkevinpal:
I don't see authModule being used anywhere in the code other than being declared here I think this is safe to remove


kevkevinpal:
@tobi-bams can you take a look to see if you can find this function used anywhere

tobi-bams:
> @tobi-bams can you take a look to see if you can find this function used anywhere

@kevkevinpal `authModule` function is not being used anywhere

PR #543, opened 2023-01-12
Feat/call recording bot

tobi-bams:


kevkevinpal (approved):
I think this is a good start, I think we might need to talk a bit about how we're getting the specific clip. Are you just looping through the specified S3 bucket to see if there's a new clip, getting the clip then notifying stakwork?

tobi-bams:
> I think this is a good start, I think we might need to talk a bit about how we're getting the specific clip. Are you just looping through the specified S3 bucket to see if there's a new clip, getting the clip then notifying stakwork?

Yes, we need to specify how we intend to get specific clips.
What was on my mind was maybe we would be using `call_id` as the name we use to store files on s3, then I keep querying s3 with that particular file name(`call_id`) till I get a response

so my request would look like this: `bucket_name.s3.amazonaws.com/folder_name/call_id.mp4`

Evanfeenstra (approved):
looks good!

src/builtin/callRecording.ts
```diff
@@ -0,0 +1,214 @@
...
+                    .setAuthor('CallRecordingBot')
+                    .setDescription('Call was recorded successfully')
+                  message.channel.send({ embed })
+                  return
+                } else {
+                  throw `Could not store in stakwork ${sendFile.status}`
+                }
+              }
+              // If recording not found after specified time then it returns an error
+              if (timeActive === 180000 && !file.ok) {
```
kevkevinpal, line 156: is this just 3 minutes?

tobi-bams: This is for test, I just updated it with my recent push

Evanfeenstra:
looks good to me!

kevkevinpal (approved):
reACK LGTM 🚀

PR #552, opened 2023-01-16
Feat/graph subscription

tobi-bams:


Evanfeenstra (commented):
what is the `chat_ids` column on the `sphinx_graph_suscription` exactly for?

I wonder if a table mapping `chat_id` to `subscription_id` might be better

tobi-bams:
> what is the `chat_ids` column on the `sphinx_graph_suscription` exactly for?
> 
> I wonder if a table mapping `chat_id` to `subscription_id` might be better

The `chat_ids` column is used to specify what tribe has access to the particular `graph_subscription`.

Also, a table mapping `chat_id` to `subscription_id` is also a good idea I guess

Evanfeenstra:
> The `chat_ids` column is used to specify what tribe has access to the particular `graph_subscription`.
> 
> Also, a table mapping `chat_id` to `subscription_id` is also a good idea I guess

The reason i ask is... Chat used to have a `contact_ids` row, then we ended up making a `ChatMembers` table, because you can add other information (role, etc). Also you can do more queries without needing to JSON.parse() the row.

kevkevinpal:
merged in https://github.com/stakwork/sphinx-relay/pull/593

PR #555, opened 2023-01-18
column for default join tribe on proxy

Evanfeenstra:


kevkevinpal:
do these tribe have to be on the proxy node itself or can they be generated by another relay? if it includes a price to join and ect

Evanfeenstra:
> do these tribe have to be on the proxy node itself or can they be generated by another relay? if it includes a price to join and ect

I was thinking just on the proxy relay itself. What do u think?

PR #556, opened 2023-01-18
Feat/search bot

tobi-bams:
Search Bot


Evanfeenstra (changes requested):
I dont like the hardcoded `3.95.131.14:5000`. Put it in the config and make that value the default config. So you dont need to add it to the JSON file. BUT it can be changed if need be.

Also what is with the hardcoded macaroon string in there?

tobi-bams:
> I dont like the hardcoded `3.95.131.14:5000`. Put it in the config and make that value the default config. So you dont need to add it to the JSON file. BUT it can be changed if need be.
> 
> Also what is with the hardcoded macaroon string in there?

Ops!!....

So sorry this is just for test's sake, for the time being, I would add the `3.95.131.14:5000` to the config

Also for the hardcoded macaroon, we need that there just to be able to test

Evanfeenstra (approved):
Ok looks good

kevkevinpal:
just a heads up we can't call jarvis-backend directly we should do it through boltwall which asks for an lsat, that port 5000 will be closed as it is insecure to keep that open. If we intend to keep up a paywall

src/utils/config.ts
```diff
@@ -134,5 +135,6 @@ export function loadConfig() {
       config.length_of_time_for_transport_token_clear ||
       DEFAULT_LENGTH_DELAY_FOR_TRANSPORT_TOKEN_DB_CLEARING,
     boltwall_server: config.boltwall_server || DEFAULT_BOLTWALL_SERVER,
+    private_graph: DEFAULT_PRIVATE_GRAPH,
```
kevkevinpal, line 138: so these private graphs should be set per tribe, but it seems like here we are just setting a single value in the config.

If I were to be in two different company tribes then this would not work

tobi-bams: Yeah!!!.....

This won't work, when I raised that with Paul he said, for now, we are believing a user belongs to only one tribe.
After this, we would focus on other complex stuffs we want to factor in, such as what you just addressed

kevkevinpal:
was merged in https://github.com/stakwork/sphinx-relay/pull/593

PR #557, opened 2023-01-19
fix: we updated to a new workflow id

kevkevinpal:
We changed the workflow id because before we were calling two seperate workflows but when they would get proccessed and sent over to jarvis-backend there would be no way to correlate the two workflows togehter since before it was all originating from stakwork but now the bot is initiating it with the call recording


tobi-bams (approved):
This looks good to me!!!
So we merged everything into one request to stakwork

PR #569, opened 2023-01-25
fix: updated the time to 60 seconds and unix time and UTC

### LINKED ISSUE #534: Create tribe bot for call recordings
Work to be done in **Relay**.

## Context:
Building the structure for private company knowledge graphs.

## Goal:
Create tribe bot which does the following:
1. detects the message that is sent when call is created from within tribe (when this message is sent, bot knows a call has started and is being recorded)
2. detects when there is an audio file that is ready to be processed (call is complete and has been stored in S3)
3. send audio file to Stakwork using stakwork endpoint and api key, and include a webhook for where to send response

## Details:
We'll need to build and use a data model for call recordings in relay, including the following:
- recording ID
- user who created the call?
- file name
- participants of the call?
- length of call?
- date/time of call
- status of call (new = created but no participants joined yet, in progress = at least 1 participant in the call, complete = all participants left the call, stored)
- webhook url for jarvis backend
- /call history

Link to project overview ticket: [https://github.com/stakwork/jarvis-backend/issues/83](url)

Image

kevkevinpal:
We had an issue of the backend giving 0 as the response for date and that is because we were not giving the correct unix timestamp, Also the videos were 30 second long so we extended it to 60 seconds

closed #534


tobi-bams (approved):

tobi-bams (approved):
This looks good to me!!!!!

PR #570, opened 2023-01-25
update: changed the query for updating total tribe message

tobi-bams:
Updated the query for updating total tribe messages


Evanfeenstra:
This query needs a `tenant` right? @tobi-bams 
```
const tribes = (await models.Chat.findAll({
        where: { ownerPubkey: contact.publicKey },
 })) as ChatRecord[]
```

Because `Chat` row is created for each tribe member. So for relay in multi-tenant mode, you need ONLY the Chats that owned by a contact on this relay. So add `tenant: contact.id` to the `where` clause

tobi-bams:
> This query needs a `tenant` right? @tobi-bams
> 
> ```
> const tribes = (await models.Chat.findAll({
>         where: { ownerPubkey: contact.publicKey },
>  })) as ChatRecord[]
> ```
> 
> Because `Chat` row is created for each tribe member. So for relay in multi-tenant mode, you need ONLY the Chats that owned by a contact on this relay. So add `tenant: contact.id` to the `where` clause

Yea, that's true.....

Thanks, lemme fix that!!

src/utils/setup.ts
```diff
@@ -157,27 +158,29 @@ const runMigrations = async () => {
...
-      for (let i = 0; i < result.length; i++) {
-        const member = result[i]
-        const totalMessages = await models.Message.count({
-          where: { sender: member.contactId, chatId: member.chatId },
-        })
-        await member.update({ totalMessages })
+    const contacts = (await models.Contact.findAll({
+      where: { isOwner: true },
+    })) as ContactRecord[]
+    for (let i = 0; i < contacts.length; i++) {
```
kevkevinpal, line 164: We can use a for each here instead of a c style loop

src/utils/setup.ts
```diff
@@ -157,27 +158,29 @@ const runMigrations = async () => {
...
+      const contact = contacts[i]
+      const tribes = (await models.Chat.findAll({
+        where: { ownerPubkey: contact.publicKey },
+      })) as ChatRecord[]
+      for (let j = 0; j < tribes.length; j++) {
```
kevkevinpal, line 169: Same here use a for each instead of c style loop

src/utils/setup.ts
```diff
@@ -157,27 +158,29 @@ const runMigrations = async () => {
...
+        const tribe = tribes[j]
+        const tribeMembers = (await models.ChatMember.findAll({
+          where: { chatId: tribe.id },
+        })) as ChatMemberRecord[]
+        for (let k = 0; k < tribeMembers.length; k++) {
```
kevkevinpal, line 174: Last one use for each instead of c style loop

kevkevinpal (commented):
For each instead of c style looping

src/utils/setup.ts
Evanfeenstra: disagree... for loop is "async aware"... forEach will be faster but could also overflow the SQLITE connection and cause a crash with many members (all queries would be running simultaneously)

PR #572, opened 2023-01-26
Feat/silent bot msg

tobi-bams:
Added the silent bot message feature


Evanfeenstra:
- in send.ts, please split out the added lines 138-164 into a separate function. No need to put that logic into the `sendMessage` function. Please try to keep code as clean as possible! So you can just add one simple if statement that calls a function.
- You need sensible defaults. The `update` command in the `callRecording` bot should ALWAYS be only_owner. That is because you dont want to expose the stakwork API key EVER. So add that default to any callRecording bot in setupHiddenBotCommands. Also, when a new callRecording ChatBot is created, automatically add `update` into the hidden commands.
- is there a test for the silent messages? It will be a good thing to add a test for

Evanfeenstra (changes requested):
https://github.com/stakwork/sphinx-relay/pull/572#issuecomment-1405835429

Evanfeenstra (approved):

PR #574, opened 2023-01-30
Fix/hide command query

tobi-bams:
Updated the hide query setup to become more flexible


tobi-bams:
> I'm kinda confused on what this is changing? Are we adding in the /callRecording bot commands here to hide them?

Oh so sorry, before now the start up script just updates the `hiddenComand` column with `update` and `hide` for the `/callRecording` bot and `hide` for every other bot

This is not a good way, because if admins have set other commands as hidden the previous code just update them to the default whenever the node restarts.

This new PR fixes that, help ensure other hidden commands specified by the admin is kept the same way even after a node is restarted...

kevkevinpal:
closing we discussed and not needed any more

PR #579, opened 2023-01-31
Modifying Content Feed Status table 

kevkevinpal:
In the migrate.ts file in utils we were using BIGINT instead of INTEGER and this was causing some versions of sqlite3 to fail when trying to migrate


kevkevinpal:
changes included in https://github.com/stakwork/sphinx-relay/pull/581

PR #581, opened 2023-02-01
feat: created the update content feed status endpoint

tobi-bams:
Summary:
This endpoint Helps update the Content Feed status


kevkevinpal (approved):
LGTM 🚀

kevkevinpal:
@Evanfeenstra one thing we changed is to use INTEGER instead of BIGINT, Tom's node didn't like when we used BIGINT for some reason. Not sure if this shouldn't be done for any specific reason because I see us using BIGINT elsewhere in the models

Evanfeenstra (approved):

Issue #584, opened 2023-02-03
Relay change that broke proxy


kevkevinpal:
This change broke proxy https://github.com/stakwork/sphinx-relay/pull/523/files

not sure why the test didn't break because users were not able to receive mqtt messages. 

We should try applying these changes and getting the tests to fail

kevkevinpal:
https://github.com/stakwork/sphinx-relay/blob/master/src/tests/controllers/tribe3Messages.test.ts

some of these tests dont even use proxy which maybe the reason this change passed integration testing

kevkevinpal:
Adding some new test changes here, will check with the change that we reverted to make sure the test fails

https://github.com/kevkevinpal/sphinx-relay/pull/21

kevkevinpal:
@Evanfeenstra looks like the cryptian issues were not just on the proxy node looks like the automated tests are failing to build the modules aswell

https://github.com/stakwork/sphinx-relay/actions/runs/4085405458/jobs/7043333139

Evanfeenstra:
@kevkevinpal you can set `jsceyptor-2` back to v0.0.1 to use mcrypt again instead of cryptian

kevkevinpal:
Ok thanks @Evanfeenstra looks like its building and passing now, going to add the change that broke proxy to the PR to get the test to fail

kevkevinpal:
This add more proxy users https://github.com/stakwork/sphinx-stack/pull/56

but it seems this wont fix the issue we had since it was trying to connect the already existing proxy users to mqtt but the way sphinx-stack is setup is that it generate new proxy users. I can explore to see if we can just have an existing state in proxy that we can startup from

PR #589, opened 2023-02-03
test: adding proxy nodes to the test case

kevkevinpal:
In this change we're adding the proxy nodes and checking if they are in the nodes file, If they do exsist then we'll try and join the tribe

This will close https://github.com/stakwork/sphinx-relay/issues/584

### DIFF
--- a/src/tests/controllers/tribe3Messages.test.ts
+++ b/src/tests/controllers/tribe3Messages.test.ts
@@ -18,6 +18,17 @@ test('test-10-tribe3Msgs: create tribe, two nodes join tribe, send messages, 2 n
 })
 
 export async function tribe3Msgs(t, node1, node2, node3) {
+  //This is checking if the proxy nodes exist and if they
+  // do then we'll use them
+  let useProxyNodes = false
+  let proxyNode1 = null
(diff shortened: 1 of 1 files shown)


kevkevinpal:
wtf it should fail for tribeMessages after this commit [ea81894](https://github.com/stakwork/sphinx-relay/pull/589/commits/ea81894e8202669c5822c6db15a14d554beab670)

kevkevinpal:
looks to be printing no alias for the proxy nodes, not sure if they're not loading in for some reason

kevkevinpal:
Only seeing one proxy user being created and put into NODES.json we should at least have two in there

Evanfeenstra:
we really should have 3 so we can do full test (proxy tribe admin, 2 proxy tribe members, one member can send to another, messages and boost sats)

kevkevinpal:
ok so this might be harder to test than I initially thought, It seems like the function we broke needs multiple users already registered to the proxy before startup. We're registering them as we go on the initialization of the test so they're still connecting

kevkevinpal:
Nonetheless we can still increase the number of proxy users since I think we only had one previously

PR #592, opened 2023-02-05
replace tv42_zbase32_gopherjs.js with zbase32.ts

antonilol:
replace an almost 70k line javascript file with my tv42/zbase32 (the lib lnd uses) compliant typescript implementation
https://github.com/antonilol/zbase32


kevkevinpal:
@antonilol So it seems like in node version 12 we get this error
`[error] 02-07-23T17:57:05 [TRIBES] error initializing TypeError: chunk.writeUintBE is not a function`

do you see a simple fix that would support node12

antonilol:
> @antonilol So it seems like in node version 12 we get this error `[error] 02-07-23T17:57:05 [TRIBES] error initializing TypeError: chunk.writeUintBE is not a function`
> 
> do you see a simple fix that would support node12

writeUintBE is an alias of writeUIntBE (note the small/capital letter i) that was added somewhere in node v12. I searched through the nodejs docs and found it in latest v12 added in v12.19.0.

Changing writeUintBE to writeUIntBE and readUintBE to readUIntBE should fix this

PR #596, opened 2023-02-09
Update/call recording retry

tobi-bams:
Summary

1. Created a new column called retry to help track the number of call retries that happen
2. Retry calls that do not have a status of `confirmed` and their `retry` value is less than 5
3. Implemented the status endpoint from stakwork, which helps us gets the status of a call recording successfully on stakwork.


Evanfeenstra (approved):

kevkevinpal (approved):
LGTM 🚀

PR #597, opened 2023-02-09
tests: adding another matrix to test node 12 and 16

kevkevinpal:
Trying to see if we can test multiple node version compatibility


kevkevinpal:
This should be good to go, we're now testing node versions 12 and 16 now

PR #602, opened 2023-02-13
Making the ports the same in Docker or Bare Metal

mishfit:
It doesn't look either the mac client win/linux or the iOS application are attempting to connect to port 3001.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Payment status: SUCCEEDED, preimage: <......>
 
 ## Network connectivity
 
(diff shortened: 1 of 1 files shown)


kevkevinpal:
So I think what might make more sense is to show the user how to change the port/validate which port to open rather than blindly using one or the other

for example we have the config file that is in `./config/app.json` which contains `node_http_protocol` `node_ip` and `node_http_port` 

where we have them being initialized in `./src/utils/config.ts`

kevkevinpal:
did you resolve your issue by changing a setting in sphinx-relay?

mishfit:
I searched for `3001` in all the client repos...nothing is trying to connect to `3001`. Swapping that out for port `3300` fixed things for me.

antonilol:
> nothing is trying to connect to `3001`.

the port is not hardcoded, it is in the connection string when connecting for the first time with the qr code or can be configured

in the sphinx relay repo you can easily search with grep, just suppress the spam from node_modules
`grep -r --exclude-dir=node_modules 3001`

kevkevinpal:
@mishfit if you take your connection string and run it through a base64 decoder you should be able to see where it's trying to connect to

mishfit:
Perhaps the issue was something else, switching to 3300 appeared to solve it for me.

PR #604, opened 2023-02-14
Feat/remove recurring call cmd

tobi-bams:
PR Summary
1. Created the list recurring command that lists the last 10 recurring calls that exist in a class or the amount the user specifies in the config
2. Made minor changes to the call recording bot


kevkevinpal (approved):
looking good minus one small nit, and I see the console.log fix aswell

LGTM 🚀

PR #605, opened 2023-02-14
config: added Env for botlwall to allow a user to set that way

kevkevinpal:


kevkevinpal:
we might need to change this in the future since this env var might be obsolete since a user might want to query multiple boltwalls. But I can add for now since it follows our config scheme

PR #611, opened 2023-02-17
Feat/block bot

tobi-bams:
PR Summary

1. Created the add command for the block bot
2. Created the remove command for the block bot
3. Intercept whenever a new user joins a tribe either kick the user out if they are on the blocked list or let them remain in the tribe


Evanfeenstra (commented):
It look good... but i wonder about listing the "kicked" members. Since you destroy the ChatMember record, it might be impossible to list them by alias.

So ... what if instead of storing an array of pubkeys in `meta`, it was just a `kicked` bool in the ChatMember table @tobi-bams ?

Evanfeenstra:
ok maybe we just dont need a "list" function. Intead just "check" if a pubkey is on the blacklist. Lets keep it the same way you have it

Aliases can be changed anyway, but pubkey are always the same for a member

Evanfeenstra (approved):

PR #613, opened 2023-02-17
Feat/kickbot test

tobi-bams:
PR Sumarry

1. Implemented test for the kick bot


Evanfeenstra (commented):
Neither the kickBot not the badgeBot tests are passing...

tobi-bams:
@Evanfeenstra, finally found a walkaround for the `kickbot` that's failing but it's not a very efficient approach. with the `setinterval` approach, some other tests were breaking.
Two possible solutions I think
1. We would have to modify tests that were failing due to the `setInterval` function
2. Try to find out why the `receiveGroupJoin` is taking time to create the user on Github, but works fine

PR #617, opened 2023-02-23
Update/only owner column

tobi-bams:
PR Summary
1. Added the `onlyOwner` column to the message table
2. update the message table if they are for only admins
3. When relaying tribe messages to new tribe member-only send messages that is not `onlyOwner`
4. Removed the hidecommand script on start-up


Evanfeenstra:
Not sure about the 
```
await models.Message.update(
          {
            onlyOwner: true,
          },
          { where: { uuid: msg.message.uuid } }
)
```

What if the Message is actually not created yet? Is the message created in the ACTION (receiveMessage)? Also does there need to be `tenant` on this query?

I wonder if there is a better way, by just adding the `onlyOwner` when the message is actually cerated, not updating it later? @tobi-bams

tobi-bams:
@Evanfeenstra for message to reach the point where we are doing the check, the message must have been created by the admin's node, because the admin is the one sending the message himself, and `uuid` is created for messages before they are added to the message table.

Yes, I should actually add `tenant` to the query.

tobi-bams:
Also, I am not sure adding `onlyOwner` at the point where a message is created is a good thing, because at that point we would have to decrypt the message, loop through all the bots active in that tribe, and get their hide commands and determine this if it's `onlyOwner` or not.
Also, this would be doing the same thing in two different parts of the codebase for the same message.

Evanfeenstra:
ok! ya just add `tenant` and it should be good i guess

PR #626, opened 2023-03-06
make mass ping much more efficient

Evanfeenstra:


src/hub.ts
```diff
@@ -156,11 +156,19 @@ async function massPingHubFromProxies(rn) {
...
+  const channelList = await Lightning.listChannels({})
+  if (!channelList) return sphinxLogger.error('failed to listChannels')
+  const { channels } = channelList
+  const localBalances = channels.map((c) => parseInt(c.local_balance))
+  const remoteBalances = channels.map((c) => parseInt(c.remote_balance))
+  const largestLocalBalance = Math.max(...localBalances)
+  const largestRemoteBalance = Math.max(...remoteBalances)
+  const totalLocalBalance = localBalances.reduce((a, b) => a + b, 0)
   await asyncForEach(owners, async (o: Contact) => {
-    const proxyNodeInfo = await proxynodeinfo(o.publicKey)
```
kevkevinpal, line 160: instead of changing all this code couldnt you just move this 2 lines up?

Evanfeenstra: well each proxynodeinfo was taking the pubkey as an argument. But i think its best just to get rid of that function since its kinda useless

PR #633, opened 2023-03-09
Feat/sentiment bot

tobi-bams:
Sentiment Bot


Evanfeenstra:
Not sure about the diff and threshold calculations in utill/sentiment.ts

- `newThreshold` should be called `newResult`
- then you need to know the "scale" of the results (are they numbers 0 to 100? 0 to 1?). So you can know the "maximum_result" number
- then the "diff" should be a percentage value: `(Math.abs(newResult - last_result) / maximum_result) * 100`. Then you only need to check that `diff >= threshold`

PR #642, opened 2023-03-23
Feat: messages deleter

tobi-bams:


Evanfeenstra:
@tobi-bams many tests are failling, any idea why?

tobi-bams:
@Evanfeenstra I guess your last change that was merged to master is what's causing the test to fail, lemme try building relay images locally and run the test and see where it's breaking

PR #652, opened 2023-04-03
fix: fixing bot message profile picture

tobi-bams:


kevkevinpal (approved):
looks good to me, and looks like its working on the test tribe

Evanfeenstra:
if we dont need a condition, then we dont need the spread operator ... {}

PR #657, opened 2023-04-13
Fix/payment error handling

tobi-bams:


Evanfeenstra:
@tobi-bams can you make a utility function like `errMsgString` instead of putting 
```
let errorMessage = ''
      if (typeof error === 'string') {
        errorMessage = error
      } else {
        errorMessage = error?.message
      }
      ```
      everywhere?

tobi-bams:
Yea....

That's true!!

Evanfeenstra:
Note from Tom:
GET /payments should NOT return failed messages.
But there should be a param to pull all payments (both failed and succeeded)

PR #658, opened 2023-04-17
update: ensured more than 10 messages are left in every chat

tobi-bams:


src/controllers/messages.ts
```diff
@@ -1002,9 +1002,30 @@ async function handleMessageDelete({
...
+    const chat_messages: { [key: string]: MessageRecord[] } = {}
+
+    for (let i = 0; i < messages.length; i++) {
+      const message = messages[i]
+      chat_messages[message.chatId] = [
+        ...(chat_messages[message.chatId] || []),
+        message,
+      ]
+    }
+    //loop through the chats and delete chat messages tht are greater than 10
```
kevkevinpal, line 1020: nit: tht -> that

kevkevinpal:
where do we check how old the message is to delete is that in another PR that we merged earlier @tobi-bams

tobi-bams:
> where do we check how old the message is to delete is that in another PR that we merged earlier @tobi-bams

Yes, that's in another PR we merged...

src/controllers/messages.ts
tobi-bams: fixed it this already

kevkevinpal:
@tobi-bams are you able to write a test for this change

maybe adds 12 messages all with the expiry date later than when we specified and then delete the messages and insure that 10 still exist

https://github.com/stakwork/sphinx-relay/blob/dff5be54aea3cdaa80b6d6275e4967b4ee8a0936/src/tests/controllers/messageLength.test.ts#L5

^ might be good to work off of that test

Evanfeenstra (commented):
Messages are always created with an ID that is higher than the previous one.

So a simpler way to do it would be to find the ID of the message that is 10th-to-last in the chat. Then you can simple to `destroy WHERE id < tenthToLastID and chat_id = CHAT_ID`

tobi-bams:
> 

Hmmm, I have been trying to wrap my head around how possible it is to get the 10th-to-last id from a chat, could you explain where, also don't forget we are also using date..

If a user sets the date to 30 days, it means if messages before 30days are more than 10 we will have to leave it

kevkevinpal:
@tobi-bams might be worth taking a look at this 
https://stackoverflow.com/questions/28504545/how-to-mock-a-constructor-like-new-date

mocking the date for the test instead of having to wait might be useful

Evanfeenstra (approved):
looks simpler and more efficient now :)

kevkevinpal (approved):
Woah nice work looks super good!

PR #661, opened 2023-04-20
update: added a fix to the boost issue

tobi-bams:


Evanfeenstra:
@tobi-bams I dont think this is the best solution. Using the global var FAILED_TO_SEND_ID could make things get complicated, like if two tribe messages are sending at the same time (from different parts of the code).

I would solve it like this:
- check if there is a contactId that is equal to the realSatsContactId
- if so, move it to the front of the array
```
const realSatsIndex = contactIds.findIndex(cid=> cid===realSatsContactId)
if (realSatsIndex>0) {
  contactIds.unshift(contactIds.splice(realSatsIndex, 1)[0])
}
```
- instead of `await asyncForEach(contactIds...`, use a `for (const contactId of contactIds) {}`
- now, instead of "return" when you want to skip a contactId (for example `if (destkey === skipPubKey)`) you can use "continue" to skip the iteration. And now if the first one fails, you can `break` so that the rest of the contactIds are not tried.

Evanfeenstra:
@tobi-bams can u write a test for failed boosts?

Where tribe admin tries to send more sats then they have. It should create a failed Message

Evanfeenstra:
Then we need to think about the next problem: tribe member boosts another member… the sats make it to the admin ok, but then the forwarded boost fails! In this case, the one who sent the original boost should see the failure, and also they should have their sats returned. Make sense?

Lets do that in a seperate PR after you write a test for the first simple case

PR #685, opened 2023-05-08
Fix/user admin boost

tobi-bams:


src/tests/configs/nodes.json
kevkevinpal, line 1: we could remove this from the diff

src/utils/reversal.ts
```diff
@@ -0,0 +1,45 @@
...
+interface ReversalInput {
+  tenant: number
+  type: string
+  errorMsg: string
+  msgUuid: string
+  chat: ChatRecord
+  sender: ContactRecord
+}
+
+export async function reversal({
```
kevkevinpal, line 15: I'm confused do why don't we send back the sats here and only in /network/send.ts

src/network/send.ts
```diff
@@ -186,6 +197,20 @@ export async function sendMessage({
...
+      where: { tenant, uuid: msg.message.uuid },
+    })) as MessageRecord
+    return await reversePayment({
+      tenant,
+      originalMessage,
+      msgToBeSent: msg,
+      error: 'user is no longer in tribe',
+      amount,
+      sender,
+    })
```
kevkevinpal, line 212: This bit of code looks pretty similar to the one below, I wonder if we could make this a function instead of repeating twice

src/controllers/messages.ts
```diff
@@ -619,6 +620,17 @@ export const receiveBoost = async (payload: Payload): Promise<void> => {
   date.setMilliseconds(0)
   if (date_string) date = new Date(date_string)
 
+  if (payload.error_message) {
```
kevkevinpal, line 623: how do we send back the sats from this point?

src/controllers/payment.ts
```diff
@@ -187,6 +188,17 @@ export const receivePayment = async (payload: Payload): Promise<void> => {
   date.setMilliseconds(0)
   if (date_string) date = new Date(date_string)
 
+  if (payload.error_message) {
```
kevkevinpal, line 191: how do we send back the sats from this point?

kevkevinpal (commented):
Looking pretty good, added some comments

src/network/send.ts
tobi-bams: Sure...

Lemme fix that...

src/utils/reversal.ts
tobi-bams: So sorry, the function name is misleading, I changed the name in my latest push

src/tests/configs/nodes.json
kevkevinpal, line 1: do we need to add this into the diff?

tobi-bams: Nope we don't

Issue #682, opened 2023-05-05
Add /dist into .gitignore


kevkevinpal:
We already have a git action to build and push to master after each push to master

https://github.com/stakwork/sphinx-relay/blob/master/.github/workflows/lint_format_and_build.yml

We might just need to modify this git action to force add something in the .gitignore but it would make our PR's not have a bunch of dist files in the diff

Evanfeenstra:
keep in mind too that the Dockerfile doesn't even run `npm run build`, it uses the /dist from the repo itself

kevkevinpal:
> keep in mind too that the Dockerfile doesn't even run `npm run build`, it uses the /dist from the repo itself

Yea I was thinking the gitaction could still build the files and commit them so they'll still be there just not when we create a PR

kevkevinpal:
not sure what is best practice tho

Evanfeenstra:
> not sure what is best practice tho

ya we are definitely not doing best practices. But its annoying for a newbie to have to understand typescript to clone and run relay. Which is why i made it this way

But if there is a way for the action to commit the files but not our PRs that would be nice...

kevkevinpal:
The above PR should close this issue, lemme know what you think. Would make the PR's easier to read

PR #686, opened 2023-05-08
feat: added dist into gitignore and force added dist folder in gitaction

kevkevinpal:
Closes https://github.com/stakwork/sphinx-relay/issues/682


.github/workflows/lint_format_and_build.yml
kevkevinpal, line 1: This file already builds and pushes the dist file on push to master I'm just adding --force because it is now in .gitignore

PR #693, opened 2023-05-22
Fix/cln signing token

tobi-bams:


Evanfeenstra (commented):
It should be 
```
const now = moment().unix()
const tsBytes = Buffer.from(now.toString(16), 'hex')
const bytesBase64 = buff.toString('base64')
const bytesUtf8 = Buffer.from(bytesBase64, 'utf8')
const sig = await LND.signBuffer(bytesUtf8, ownerPubkey)
```

Evanfeenstra:
can u add to tribe create test... check that the owner_pubkey equals your own pubkey after the tribe is created

.github/workflows/integration_test.yml
```diff
@@ -66,21 +66,22 @@ jobs:
           npm install && npm run build && docker build -t sphinxlightning/sphinx-relay .
       - name: Checkout stack
         run: |
-          git clone https://github.com/stakwork/sphinx-stack.git stack
+          git clone -b feat/cln-support https://github.com/stakwork/sphinx-stack.git stack
```
kevkevinpal, line 69: just adding a comment here to remember to merge stack before merging this PR

.github/workflows/integration_test.yml
```diff
@@ -66,21 +66,22 @@ jobs:
...
           chmod 777 -R ./proxy;
+          chmod 777 -R ./cln;
       - name: Check for NODES.json
         uses: nick-fields/retry@v2
         with:
           timeout_minutes: 10
           max_attempts: 3
           command: |
-            GITACTION_ENV=gitactionenv docker-compose -f ./stack/alts/proxy.yml --project-dir ./stack up -d;
+            GITACTION_ENV=gitactionenv docker-compose -f ./stack/alts/cln-proxy.yml --project-dir ./stack up -d;
```
kevkevinpal, line 84: just a question

So is the current test environment 3 LND nodes and 1 proxy node connected to CLN?

Evanfeenstra: yes. but for automated tests we should have both LND+Proxy and CLN+Proxy setups (there are two different compose files)

PR #709, opened 2023-06-14
update: ensured when a test fails for proxy node, the route actually …

tobi-bams:
…fails


.github/workflows/integration_test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           npm install && npm run build && docker build -t sphinxlightning/sphinx-relay .
       - name: Checkout stack
         run: |
-          git clone -b update/proxy-tag https://github.com/stakwork/sphinx-stack.git stack
+          git clone  https://github.com/stakwork/sphinx-stack.git stack
```
kevkevinpal, line 69: didn't realize we were still using a different branch we should get this updated to master

PR #712, opened 2023-06-20
feat: send message directly if users co-exist on the same relay

tobi-bams:


kevkevinpal (approved):
lgtm just needs to be rebased

PR #714, opened 2023-06-22
fix resub mqtt on reconnect

Evanfeenstra:


src/utils/tribes.ts
```diff
@@ -112,6 +112,9 @@ async function initializeClient(
           sphinxLogger.info(`connected!`, logging.Tribes)
           if (!clients[username]) clients[username] = {}
           clients[username][host] = cl // ADD TO MAIN STATE
+          // subscribe here
+          await specialSubscribe(cl, contact)
```
kevkevinpal, line 116: The only thing I can think of is this is in a scope withing cl.on and it should be outside?

Evanfeenstra:
@tobi-bams maybe tomorrow you can try running this locally and see what the issue is?

kevkevinpal:
just pushed [4c7f49c](https://github.com/stakwork/sphinx-relay/pull/714/commits/4c7f49c470e2e92ce29358d4fcc728f5cc1bee7d) trying to see if taking the subscribe out of the anon function would help it connect properly

kevkevinpal:
I have a feeling that client.on connect can be still waiting while the subscribe happens so that might cause it to fail but I'll see what the tests say

kevkevinpal:
I can try building a docker image and running this on my node to see if mqtt reconnects properly

kevkevinpal:
nvm not sure if this fixes issue

kevkevinpal:
I ran this on my raspi
`rfkill block wifi && sleep 10 && rfkill unblock wifi`
I saw the `CLOSE` log and the `reconnect!` log

then had someone send a message in a tribe where I was not the owner and the owner was not sending the message and was able to see it.

So I think this is reconnecting properly now!

PR #715, opened 2023-06-26
Feat/swarm invites

tobi-bams:


src/grpc/lightning.ts
```diff
@@ -1257,7 +1257,7 @@ export async function getInvoiceHandler(
                 preimage: response?.settled
                   ? response?.r_preimage.toString('hex')
                   : '',
-                amount: convertMsatToSat(response.amt_paid),
```
kevkevinpal, line 1260: I'm confused why this change is needed for the swarm invites change?

PR #722, opened 2023-07-17
cleanup imports with new eslint rule to keep imports organized

kevkevinpal:
Added new eslint rule to keep imports in order, I find this useful as it'll help the person coding organize in their head easier what they're importing in the file


kevkevinpal:
I need to remove the yarn.lock from this PR

PR #733, opened 2023-07-28
added thread_uuid to the media attachment endpoint

kevkevinpal:
opening this PR but I still want to add a test for the thread_uuid on attachemnts


kevkevinpal:
added a test in this commit not exactly sure if it will work

[f31dede](https://github.com/stakwork/sphinx-relay/pull/733/commits/f31dedeea9df03858c062feb01c3651443b3e451)

PR #743, opened 2023-08-02
Feat/spam gone bot

tobi-bams:


tobi-bams:
@kevkevinpal, please when updating relays for our nodes, please run npm install because I updated the sphinx-bot package. Just incase you run into errors

PR #748, opened 2023-08-04
ML bot

tobi-bams:


src/builtin/ml.ts
```diff
@@ -0,0 +1,225 @@
...
+          message: message.content,
+          webhook: `${host_name}/ml`,
+        }),
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/json',
+        },
+      })
+      const j = await r.json()
+      console.log('ML body j', j)
```
kevkevinpal, line 173: we can remove this log

src/builtin/ml.ts
```diff
@@ -0,0 +1,225 @@
...
+      if (!process_id) {
+        const embed = new Sphinx.MessageEmbed()
+          .setAuthor('ML Bot')
+          .setDescription('failed to process message')
+          .setOnlyUser(parseInt(message.member.id || '0'))
+        message.channel.send({ embed })
+        return
+      }
+
+      console.log('PROCESS ID!!!', process_id)
```
kevkevinpal, line 192: we can remove this log

src/builtin/ml.ts
```diff
@@ -0,0 +1,225 @@
...
+            await bot.update({ meta: JSON.stringify(meta) })
+            await botResponse(
+              ML_BOTNAME,
+              'URL updated successfully',
+              ML_PREFIX,
+              tribe.id,
+              message,
+              cmd
+            )
+            return
```
kevkevinpal, line 75: for each of these switch statements it would be much cleaner if we just had a function for each one
like
```
case 'url':
   await url()
case 'api_key':
   await api_key()
case 'kind':
   await kind()
default:
   defaultFunc()
```

src/builtin/ml.ts
```diff
@@ -0,0 +1,225 @@
...
+          .setOnlyUser(parseInt(message.member.id || '0'))
+        message.channel.send({ embed })
+        return
+      }
+
+      let host_name = config.host_name
+      if (!host_name.startsWith('http')) {
+        host_name = `https://${host_name}`
+      }
+      console.log('ml bot hostname', host_name)
```
kevkevinpal, line 160: dont think we need this console.log

src/controllers/botapi/index.ts
```diff
@@ -325,3 +327,25 @@ export async function validateAction(a: Action): Promise<ChatAndOwner | void> {
   })) as ContactRecord
   return { chat: theChat, owner }
 }
+
+export async function processMlCallback(req: Req, res: Res): Promise<void> {
+  console.log('ML BODY', req.body)
```
kevkevinpal, line 332: can remove this log

src/network/send.ts
```diff
@@ -148,7 +149,7 @@ export async function sendMessage({
         sphinxLogger.info(`[Network] => isBotMsg`, logging.Network)
         // return // DO NOT FORWARD TO TRIBE, forwarded to bot instead?
       }
-      if (msg.sender.role === constants.chat_roles.owner && msg.type === 0) {
+      if (msg.type === 0) {
```
kevkevinpal, line 152: I'm lost on why we removed this check?

kevkevinpal (commented):
added some comments mostly pertaining to cleaning up the code a bit

src/network/send.ts
tobi-bams: I removed it because I am already checking to ensure only the tribe owner can execute this block of code: https://github.com/stakwork/sphinx-relay/blob/ml-bot/src/network/send.ts#L133

src/builtin/ml.ts
```diff
@@ -0,0 +1,161 @@
...
+      setTimeout(() => {
+        delete CALLBACKS[process_id]
+      }, 5 * 60 * 1000)
+    } catch (e) {
+      sphinxLogger.error(`ML CALL FAILED: ${e}`, logging.Bots)
+    }
+  })
+}
+
+export async function sleep(ms: number): Promise<void> {
```
kevkevinpal, line 159: Are you able to import sleep from helpers.ts?

PR #752, opened 2023-08-11
added updat/change-example-bot branch to stack

tobi-bams:


.github/workflows/cln_proxy_integration_test.yml
```diff
@@ -99,6 +99,10 @@ jobs:
         with:
           source: 'stack/relay/NODES.json'
           target: 'relay/src/tests/configs/nodes.json'
+      - name: Sleep for 1 minutes
+        uses: jakejarvis/wait-action@master
+        with:
+          time: '1m'
```
kevkevinpal, line 105: why are we sleeping here can't we add another sleep in the check for NODES.json step?

.github/workflows/cln_proxy_integration_test.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           npm install && npm run build && docker build -t sphinxlightning/sphinx-relay .
       - name: Checkout stack
         run: |
-          git clone https://github.com/stakwork/sphinx-stack.git stack
+          git clone -b updat/change-example-bot https://github.com/stakwork/sphinx-stack.git stack
```
kevkevinpal, line 73: just need to change this back before merging

.github/workflows/cln_proxy_integration_test.yml
tobi-bams: Bot setup only works after relay_setup has exited, so I am waiting 1 minute after relay_setup has exited for both bot containers to come up

PR #753, opened 2023-08-23
Reorganized configs to be in categories

kevkevinpal:
I think this will help people setting up their sphinx relay the first time cause before reading through the configs was really confusing


Evanfeenstra:
LGTM

kevkevinpal:
not sure why but 8 of the tests keep failing

PR #773, opened 2023-09-25
Fix/socket hang

tobi-bams:
Added some seconds sleep after tribe member leaves tribe before tribe admin deletes tribe to prevent socket hang


kevkevinpal:
lgtm will help our test suite

Issue #774, opened 2023-10-09
Endpoint for marking messages as read/unread


uncleJim21:
@Evanfeenstra @kevkevinpal we are responding to this request for ability to mark messages as read/unread to achieve parity with other similar apps: https://github.com/stakwork/sphinx-ios/issues/305. 

We have the app side database syncing but it occurred to us that we need to likewise sync up the relay with the changes as well. I'm thinking a simple PUT or PATCH http endpoint that just takes the message & chat UIDs would be sufficient.

Let me know your thoughts.

tomastiminskas:
I would make the endpoint receives `chat_id` and array of `message_ids` to be set as `seen = false`. That way we can send the chat id and the last message id to be set as unseen as we are doing on the client side.

uncleJim21:
On second thought the most effective way would be to have the following parameters:
chatID
shouldMarkUnread

if the flag is true it just marks the last message in the chat as unread.

If it’s false it marks all messages as read.

Those are the only two possible use cases for this endpoint.

the client shouldn’t have to process all the messageIds I think that’s over complicated.

tomastiminskas:
@uncleJim21 we already have:

 `POST /messages/\(chatId)/read` 

so I would make a new one for unread:

 `POST /messages/\(chatId)/unread` 
 
 This new one would set chat and last message of the chat as `seen = false` (as you suggested)
 
 @kevkevinpal @Evanfeenstra could any of you take a look at this? So Jim can finish the feature. It should be a simple addition. And you can deploy to my nodes, so Jim can test on Tom Sim account

uncleJim21:
@tomastiminskas I took a similar approach that just takes a flag and either calls the function you mentioned or it implements the mark as unread feature

tomastiminskas:
@uncleJim21 that would work. Maybe let's Evan or Kev decide the best way to do it and inform us. I thought about a new one because the one we have for read can't be just replaced for the flag one since we need to keep it alive for old versions of the app using it

uncleJim21:
@tomastiminskas agreed we need backward compatibility. I suppose your way is more consistent with the syntax. My code can refactor pretty easily to do it your way

tomastiminskas:
@Evanfeenstra @kevkevinpal any updates on this? Thanks

Evanfeenstra:
not sure why the push notification is sent, after that is removed we can merge

PR #776, opened 2023-10-10
add endpoint that can mark a given chat as read/unread

uncleJim21:
attempting to address #774


uncleJim21:
@kevkevinpal @Evanfeenstra I took a stab at this and I think I am close but I do not have a great way to test this. Could you recommend a way to set up a test environment so I can prove out this does what I expect? Or if it's more convenient just hop on a call with me while you review so I can check out your setup

kevkevinpal:
@uncleJim21 I pushed a new test can you check to see if this looks good to you, it tests for that `seen` field in the chat

tomastiminskas:
Looks ok to me. If @uncleJim21 agrees we can push to my 2 nodes so Jim can test on Tom Sim. Thanks

src/controllers/messages.ts
```diff
@@ -797,6 +797,72 @@ export const receiveDeleteMessage = async (payload: Payload): Promise<void> => {
...
+    }
+    //mark the chat as unseen
+    const chat: Chat = (await models.Chat.findOne({
+      where: { id: chat_id, tenant },
+    })) as Chat
+    if (chat) {
+      resetNotifyTribeCount(parseInt(chat_id))
+      await chat.update({ seen: false })
+      success(res, {})
+      sendNotification(chat, '', 'badge', owner)
```
kevkevinpal, line 849: why are we sending a push notification here? Weird to send a notification for something we do manually

uncleJim21: good point. I went ahead and removed that line.

tomastiminskas:
@kevkevinpal can you confirm if this is deployed to Tom Sim node? 

Once you confirm @uncleJim21 can finish the feature testing there

PR #781, opened 2023-10-20
update: added help to hidden commands

tobi-bams:


tobi-bams:
@tomsmith8 please any tribe that has the `spam_gone` bot should be uninstalled and installed again for this to work.

PR #815, opened 2026-06-14
Fail fast when integration test environment is unavailable

### LINKED ISSUE #415: Intergration tests stalling
Sometimes the integration tests stalls if the docker env doesn't startup properly

We can solve this two ways 
1). fail fast so rather than stalling notice that the env didn't start up correctly and just fail
2). restart the test if the env didn't startup properly

we could do 1 first and then do 2

caydyan:
## Summary
- add a preflight check before AVA integration tests that verifies every configured relay responds to /contacts
- add default HTTP timeouts for shared integration test request helpers
- document the readiness and per-request timeout environment variables

Fixes #415

Bounty payout BTC address: bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc

## Validation
- npx tsc --noEmit --pretty false
- git diff --check
- SPHINX_TEST_ENV_TIMEOUT_MS=500 npm test (expected fail-fast locally because relay ports 3001-3004 are not running; the command exits before AVA and reports ECONNREFUSED for the configured nodes)


caydyan:
Current-head validation rerun on `92bae684eaf91569e50fea3729bd9cf8f628acd2`:

- `npx tsc --noEmit --pretty false` passed.
- `git diff --check origin/master...HEAD` passed.
- `SPHINX_TEST_ENV_TIMEOUT_MS=500 npm test` exercised the new preflight and failed fast before AVA because the local integration relay nodes are unavailable (`bob`/`alice`/`carol`/`dave`/virtual nodes all returned `ECONNREFUSED`). That is the intended local smoke behavior for this branch: it reports the missing environment immediately instead of stalling in the integration suite.

Local worktree is clean after the rerun.

caydyan:
Hi, just checking in on this after the validation rerun above.

Current status from my side:

- PR #815 implements fail-fast behavior for stalled integration-test relay startup.
- `npx tsc --noEmit --pretty false` passed on head `92bae684eaf91569e50fea3729bd9cf8f628acd2`.
- `git diff --check origin/master...HEAD` passed.
- The local smoke path now exits with explicit unavailable relay-node diagnostics instead of hanging indefinitely.

If this satisfies issue #415's 300,000 sats bounty, BTC payout address is:

`bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

Happy to rerun validation or adjust the behavior if you want a different timeout/error shape.

